PR #6, opened 2018-12-05
Specify versions instead of revisions

benjigoldberg:
Use versions instead of specific revisions


Gopkg.toml
```diff
@@ -1,6 +1,6 @@
 [[constraint]]
   name = "github.com/Shopify/sarama"
-  revision = "e7238b119b7daab993720f0153eafb88e2b0ac1f"
+  version = "~1.19.0"
```
briancharous: oh yeah i forgot that we had to roll in the revision because of some patch that wasn't in a cut release yet

PR #7, opened 2018-12-06
add simple environmental variable lookup with default

spotrad:
Super simple environmental variable lookup with default value.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -65,3 +65,7 @@
 [prune]
   go-tests = true
   unused-packages = true
+
+[[constraint]]
+  name = "github.com/google/uuid"
+  version = "1.1.0"
--- a/utils.go
+++ b/utils.go
@@ -1,0 +1,28 @@
+// Copyright 2018 SpotHero
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tools
+
+import (
+	"os"
+)
+
+// GetEnvVal takes in an environmental variable key (key) and a default value (defaultVal)
+// and returns the value of the environmental variable key if found or the defualt value provided.
+func GetEnvVal(key string, defaultVal string) string {
+	val, ok := os.LookupEnv(key); if ok{
+		return val
+	}
+	return defaultVal
+}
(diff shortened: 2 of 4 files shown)


utils.go
```diff
@@ -0,0 +1,28 @@
...
+
+package tools
+
+import (
+	"os"
+)
+
+// GetEnvVal takes in an environmental variable key (key) and a default value (defaultVal)
+// and returns the value of the environmental variable key if found or the defualt value provided.
+func GetEnvVal(key string, defaultVal string) string {
```
benjigoldberg: @spotrad we usually use [Cobra](https://github.com/spf13/cobra) and [Viper](https://github.com/spf13/viper) to create a CLI that also works with environment variables. We typically bind those environment variables to our Cobra settings with this functionality in [`cli.go`](https://github.com/spothero/tools/blob/master/cli.go#L39). This function exposes every CLI [pflag](https://godoc.org/github.com/spf13/pflag) surfaced as an app-name prefixed environment variable, eg if you specified a flag  [like this example for HTTP address](https://github.com/spothero/tools/blob/master/cli.go#L61) and your app was named [`example_server`](https://github.com/spothero/tools/blob/master/examples/example_server.go#L38), you could specify the flag at runtime in the following ways:

1. `./example_server --address localhost`
2. `EXAMPLE_SERVER_ADDRESS=localhost ./example_server`

Would this satisfy your needs, or are you looking to grab other kinds of environment variables?

utils_test.go
```diff
@@ -0,0 +1,39 @@
...
+	"os"
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetEnvVal(t *testing.T) {
+	defaultVal := "FAKE"
+	key := uuid.New().String()
```
benjigoldberg: Is it possible to just pick a fake value as you have above and drop the dependency on UUID?

Gopkg.toml
```diff
@@ -65,3 +65,7 @@
 [prune]
   go-tests = true
   unused-packages = true
+
+[[constraint]]
+  name = "github.com/google/uuid"
```
benjigoldberg: I imagine someday this will make its way into this package, but at this point since its only used in the test I would like to avoid the additional dependency unless necessary.

utils.go
benjigoldberg: Yeah I thought about this a bit and chatted it over w/ @briancharous -- I hate to reject a PR because we really appreciate and value the contribution, but I'm having trouble figuring out where this lives for now. Most of the functionality in `tools` is grouped into bundles of substantial and related functionality (http servers, kafka, sentry, etc). I tend to think having `tools/utils.go` could be a bit of a slipper slope of one-offs that people arent aware of. What do you think? If we can identify a discrete group of functionality where this belongs and rename the file to match that, I think I'd still be open to the PR.

spotrad: I am cool with that. I've seen `uitls` files go well or completely off the rails. Not worth the risk with something this small. I'll kill this PR

PR #11, opened 2018-12-10
Remove Viper from CLI registration commands

benjigoldberg:
Removes Viper from the name of pflag registration commands


zwagner86:
Why would we remove something as badass-sounding as Viper?!

benjigoldberg:
🐍

PR #19, opened 2019-03-14
Add method to mock consumer to emit offsets

briancharous:
Can be used to simulate kafka consumer exit behavior

### DIFF
--- a/kafka_mock.go
+++ b/kafka_mock.go
@@ -2,34 +2,43 @@ package tools
 
 import (
 	"context"
-	"github.com/stretchr/testify/mock"
 	"sync"
+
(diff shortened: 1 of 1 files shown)


kafka_mock.go
```diff
@@ -2,34 +2,43 @@ package tools
...
 import (
 	"context"
-	"github.com/stretchr/testify/mock"
 	"sync"
+
+	"github.com/stretchr/testify/mock"
 )
 
 // MockKafkaConsumer implements KafkaConsumerIface for testing purposes
 type MockKafkaConsumer struct {
```
greg-mclain: The `kK` messes with my head. At first glance I think it's a typo. 😛

PR #20, opened 2019-03-14
Remove logging sentry config from server

briancharous:
There's no reason that the http server be attached to tracing or logging config; the user of tools should be able to decide how and when they want to create the logger and tracer.


briancharous:
@greg-mclain The server is still instrumented because it still captures prometheus metrics and provides pprof endpoints

PR #25, opened 2019-04-05
cleanup makefile with lint and fix linter issues

benjigoldberg:
Fixes linting issues and overhauls makefile to our newer standards.

### DIFF
--- a/logging.go
+++ b/logging.go
@@ -64,6 +64,9 @@ func (lc *LoggingConfig) InitializeLogger() {
 		logConfig.EncoderConfig.EncodeLevel = zapcore.CapitalColorLevelEncoder
 		logConfig.Level = zap.NewAtomicLevelAt(level)
 		Logger, err = zap.NewDevelopment()
+		if err != nil {
+			fmt.Printf("error creating logger configuration: %s\n", err.Error())
+		}
 	} else {
 		logConfig = zap.Config{
 			Level:             zap.NewAtomicLevelAt(level),
@@ -82,7 +85,7 @@ func (lc *LoggingConfig) InitializeLogger() {
 	}
 	Logger, err = logConfig.Build()
 	if err != nil {
-		fmt.Printf("Error initializing Logger: %s\n", err.Error())
+		fmt.Printf("error initializing Logger: %s\n", err.Error())
 	}
 	if lc.SentryLoggingEnabled && err == nil {
 		// attach Sentry core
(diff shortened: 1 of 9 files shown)


sentry.go
```diff
@@ -46,11 +46,17 @@ func (sc *SentryConfig) InitializeRaven() {
 // SentryCore Implements a zapcore.Core that sends logged errors to Sentry
 type SentryCore struct {
 	zapcore.LevelEnabler
+	withFields []zapcore.Field
 }
 
 // With adds structured context to the Sentry Core
 func (c *SentryCore) With(fields []zapcore.Field) zapcore.Core {
-	return c.With(fields)
```
benjigoldberg: Linter was complaining that this would cause infinite recursion -- which it absolutely would. Since we're implementing the zapcore interface, we aren't actually calling the underlying `.With` from the `Core` here. This is based on what `zap.Logger` does: https://github.com/uber-go/zap/blob/master/logger.go#L159

and then we append any `withFields` at the core level to the `Write` operation on [line 94](https://github.com/spothero/tools/pull/25/files#diff-0b771ec7e8ca759d819dc583bbdc795aR94).

tracing.go
```diff
@@ -56,13 +56,14 @@ func (tc *TracingConfig) ConfigureTracer() io.Closer {
...
-		Sampler:  &samplerConfig,
-		Reporter: &reporterConfig,
-		Disabled: !tc.Enabled,
+		ServiceName: tc.ServiceName,
+		Sampler:     &samplerConfig,
+		Reporter:    &reporterConfig,
+		Disabled:    !tc.Enabled,
 	}
 
-	tracer, closer, err := jaegerConfig.New(
```
benjigoldberg: `New` was deprecated. `ServiceName` now comes from `Configuration`

logging.go
```diff
@@ -64,6 +64,9 @@ func (lc *LoggingConfig) InitializeLogger() {
 		logConfig.EncoderConfig.EncodeLevel = zapcore.CapitalColorLevelEncoder
 		logConfig.Level = zap.NewAtomicLevelAt(level)
 		Logger, err = zap.NewDevelopment()
+		if err != nil {
+			fmt.Printf("error creating logger configuration: %s\n", err.Error())
```
briancharous: Should be `fmt.Errorf` IMO

logging.go
```diff
@@ -82,7 +85,7 @@ func (lc *LoggingConfig) InitializeLogger() {
 	}
 	Logger, err = logConfig.Build()
 	if err != nil {
-		fmt.Printf("Error initializing Logger: %s\n", err.Error())
+		fmt.Printf("error initializing Logger: %s\n", err.Error())
```
briancharous: same comment as above

sentry.go
briancharous: Ha good catch

PR #26, opened 2019-04-15
Feature/logging updates

benjigoldberg:
Adds a new and improved `log` module. Will phase out `logging` over the next few days.


benjigoldberg:
/bump minor -- we're still pre-1.0 so I'm not going to rev major since the API isnt yet declared stabled.

log/log.go
```diff
@@ -0,0 +1,130 @@
...
+	} else {
+		logConfig = zap.Config{
+			Level:             zap.NewAtomicLevelAt(level),
+			Development:       false,
+			DisableStacktrace: false,
+			Sampling: &zap.SamplingConfig{
+				Initial:    lc.SamplingInitial,
+				Thereafter: lc.SamplingThereafter,
+			},
+			Encoding:         "json",
```
briancharous: Should this be configurable but default to json?

log/log.go
```diff
@@ -0,0 +1,130 @@
...
+			EncoderConfig:    zap.NewProductionEncoderConfig(),
+			OutputPaths:      append(lc.OutputPaths, "stdout"),
+			ErrorOutputPaths: append(lc.ErrorOutputPaths, "stderr"),
+			InitialFields:    map[string]interface{}{"appVersion": lc.AppVersion, "gitSha": lc.GitSha},
```
briancharous: I feel like `AppVersion` and `GitSha` should also be optional

log/log.go
```diff
@@ -0,0 +1,130 @@
...
+// useful if you wish for all downstream logs from the site of a given context to include some
+// contextual information. For example, once your application has unpacked the Trace ID, you may
+// wish to log that information with every future request.
+func NewContext(ctx context.Context, fields ...zapcore.Field) context.Context {
+	return context.WithValue(ctx, logKey, Get(ctx).With(fields...))
+}
+
+// Get returns the logger wrapped with the given context. This function is intended to be
+// used as a mechanism for adding scoped arbitrary logging information to the logger. If a nil
+// context is passed, the default global logger is returned.
```
briancharous: Hmm... I'd be in favor of killing the global logger altogether

benjigoldberg: Why? I think this works nicely and you don't necessarily always have a context to make use of.

benjigoldberg: Plus, the hidden logger forces you to use the WithContext interface, so in theory you _should_ always be doing the appropriate thing

log/log.go
benjigoldberg: Yes, agreed. Optional and supplied as config 👍

log/log.go
benjigoldberg: agreed, good idea

log/log.go
```diff
@@ -0,0 +1,136 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
+
+// logKey is the type used to uniquely place the logger within context.Context
```
greg-mclain: nitpick, but with the addition of `ctxKey` you may want to either move this comment down a line, or change it from `logKey` to `ctxKey`. We probably don't need a comment on both fields

benjigoldberg: nice catch

log/log.go
```diff
@@ -0,0 +1,137 @@
...
+
+// metricsHook is a callback hook used to track logging metrics at runtime
+func (lc *LoggingConfig) metricsHook(entry zapcore.Entry) error {
+	lc.counter.With(prometheus.Labels{"level": entry.Level.CapitalString()}).Inc()
+	return nil
+}
+
+// InitializeLogger sets up the logger. This function should be called as soon
+// as possible. Any use of the logger provided by this package will be a nop
+// until this function is called
```
wilsoniya: period

benjigoldberg: 👍

log/log.go
```diff
@@ -0,0 +1,135 @@
...
+
+// logKey is the value used to uniquely place the logger within context.Context
+const logKey ctxKey = iota
+
+// logger is the default zap logger
+var logger = zap.NewNop()
+
+// LoggingConfig defines the necessary configuration for instantiating a Logger
+type LoggingConfig struct {
+	UseDevelopmentLogger bool                   // If true, use default zap development logger settings
```
wilsoniya: beautiful

log/log.go
```diff
@@ -0,0 +1,135 @@
...
+	SamplingThereafter   int                    // Sampling rate after that initial cap is hit, per cycle
+	Encoding             string                 // One of "json" or "console"
+	Fields               map[string]interface{} // These fields will be applied to all logs. Strongly recommend `version` at a minimum.
+	Cores                []zapcore.Core         // Provides optional functionality to allow users to Tee log output to additional Log cores
+	Options              []zap.Option           // Users may specify any additional zap logger options which override defaults
+	counter              *prometheus.CounterVec // Counter tracks the number of logs by level
+}
+
+// metricsHook is a callback hook used to track logging metrics at runtime
+func (lc *LoggingConfig) metricsHook(entry zapcore.Entry) error {
```
briancharous: This is pretty nitpicky but I don't think that these receivers need to be pointers

log/log_test.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			assert.Equal(t, test.expectedOutcome, NewContext(test.ctx, test.fields...))
+		})
+	}
+}
+
+func TestMetricsHook(t *testing.T) {
+	tests := []struct {
```
briancharous: There's only 1 test case here and the code its testing is like 2 lines long without any conditional branches, personally I'd get rid of the subtest and just write a single unit test in here.

benjigoldberg: yeah I know, I was being lazy b/c I set this up thinking I'd have more tests. Then, I decided I already set it up so I could just leave it for the next person.

log/log.go
briancharous: Yeah I guess this is fine. The alternative would be either erroring out or silently not logging, both of which are worse than a global.

PR #27, opened 2019-04-16
http refactor

benjigoldberg:
WIP @briancharous can you give some early feedback?


http/metrics.go
```diff
@@ -0,0 +1,93 @@
...
+		case sr.StatusCode >= http.StatusMultipleChoices:
+			statusClass = "3xx"
+		case sr.StatusCode >= http.StatusOK:
+			statusClass = "2xx"
+		default:
+			statusClass = "1xx"
+		}
+		labels := prometheus.Labels{
+			"path":         r.URL.Path,
+			"status_class": statusClass,
```
briancharous: If the actual status code is being captured, is this necessary? The status code class can be figured out later in the prometheus query.

http/middleware.go
```diff
@@ -0,0 +1,66 @@
...
+type StatusRecorder struct {
+	http.ResponseWriter
+	StatusCode int
+}
+
+// WriteHeader implements the http ResponseWriter WriteHeader interface. This function acts as a
+// middleware which captures the StatusCode on the StatusRecorder and then delegates the actual
+// work of writing the header to the underlying http ResponseWriter.
+func (sr StatusRecorder) WriteHeader(code int) {
+	sr.StatusCode = code
```
briancharous: receiver needs to be a pointer

http/middleware.go
```diff
@@ -0,0 +1,66 @@
...
+
+// MiddlewareFunc defines a middleware function used in processing HTTP Requests. Request
+// preprocessing may be specified in the body of the middleware function call. If post-processing
+// is required, please use the returned deferable func() to encapsulate that logic.
+type MiddlewareFunc func(*StatusRecorder, *http.Request) (func(), *http.Request)
+
+// middleware defines a collection of middleware functions. All functions must return a deferable
+// call, whether or not the callback performs any work in addition to the http.Request.
+type middleware struct {
+	functions []MiddlewareFunc
```
briancharous: could be `type middleware []MiddlewareFunc` instead of a struct

http/middleware.go
```diff
@@ -0,0 +1,150 @@
...
+//
+// The returned HTTP Request includes the wrapped OpenTracing Span Context.
+func TracingMiddleware(sr *StatusRecorder, r *http.Request) (func(), *http.Request) {
+	wireContext, err := opentracing.GlobalTracer().Extract(
+		opentracing.HTTPHeaders,
+		opentracing.HTTPHeadersCarrier(r.Header))
+	if err != nil {
+		log.Get(r.Context()).Debug("Failed to extract opentracing context on an incoming HTTP request.")
+	}
+	span, spanCtx := opentracing.StartSpanFromContext(r.Context(), r.URL.Path, ext.RPCServerOption(wireContext))
```
greg-mclain: Same comment as above, we'll want to use the path template (unless I'm misunderstanding and `r.URL.Path` IS the path template, which is quite possible 🙂 ).

http/metrics.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+type Metrics struct {
+	serverName string
+	counter    *prometheus.CounterVec
+	duration   *prometheus.HistogramVec
+}
+
+func NewMetrics(serverName string) Metrics {
```
briancharous: It might be nice to allow users of this to pass in a prometheus registry in case they don't want to use the global one. It might also make tests a little cleaner.

http/metrics.go
```diff
@@ -0,0 +1,78 @@
...
+			Help: "Total number of HTTP Requests received",
+		},
+		[]string{
+			// The path recording the request
+			"path",
+			// The Specific HTTP Status Code
+			"status_code",
+		},
+	)
+	prometheus.MustRegister(histogram)
```
briancharous: Might be better to allow the user of this function to decide if they want to die if prometheus errors instead of panicking here.

benjigoldberg: @briancharous take a look at how I solved this and lmk what you think. Passing back an error from this fn is kind of a pain in that it makes the upstream handling a lot more complex.

benjigoldberg:
Also I'm sorry this got so huge. That wasn't my original intent but I ended up writing a _lot_ of tests.

http/metrics_test.go
```diff
@@ -19,32 +19,72 @@ import (
...
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			registry := prometheus.NewRegistry()
+			metrics := NewMetrics("test", registry, test.mustRegister)
+			if test.duplicate {
+				if test.mustRegister {
+					assert.Panics(t, func() { NewMetrics("test2", registry, test.mustRegister) })
+				} else {
+					_ = NewMetrics("test2", registry, test.mustRegister)
```
briancharous: Wrap this in `assert.NotPanics` so in case some change results in a panic the whole test suite doesn't bomb out and it's more clear what test failed.

http/utils/utils.go
```diff
@@ -0,0 +1,46 @@
+// Copyright 2019 SpotHero
```
briancharous: I assume that you put this in a separate package to avoid circular imports, but If you think it belongs in a separate package, naming a package `utils` is [bad practice](https://blog.golang.org/package-names). You probably considered this but could the logging middleware live in the `http` package?

log/middleware.go
```diff
@@ -0,0 +1,48 @@
...
+// On outbound response return these attributes include all of the above as well as:
+// * HTTP response code
+func LoggingMiddleware(sr *utils.StatusRecorder, r *http.Request) (func(), *http.Request) {
+	remoteAddress := zap.String("remote_address", r.RemoteAddr)
+	method := zap.String("http_method", r.Method)
+	path := zap.String("path", utils.FetchRoutePathTemplate(r))
+	query := zap.String("query_string", r.URL.Query().Encode())
+	hostname := zap.String("hostname", r.URL.Hostname())
+	port := zap.String("port", r.URL.Port())
+	Get(r.Context()).Info("Request Received", remoteAddress, method, path, query, hostname, port)
```
briancharous: I thought we were standardizing on lowercase log messages

log/middleware.go
```diff
@@ -0,0 +1,48 @@
...
+//  On inbound request received these attributes include:
+// * The remote address of the client
+// * The HTTP Method utilized
+// * The hostname specified on this request
+// * The port specified on this request
+//
+// On outbound response return these attributes include all of the above as well as:
+// * HTTP response code
+func LoggingMiddleware(sr *utils.StatusRecorder, r *http.Request) (func(), *http.Request) {
+	remoteAddress := zap.String("remote_address", r.RemoteAddr)
```
briancharous: Is this the IP of the sender? This might not be useful because most of the time it's going to be the IP of the kubernetes ingress, no?

tracing.go
```diff
@@ -84,3 +88,47 @@ func TraceOutbound(r *http.Request, span opentracing.Span) {
...
+	wireContext, err := opentracing.GlobalTracer().Extract(
+		opentracing.HTTPHeaders,
+		opentracing.HTTPHeadersCarrier(r.Header))
+	if err != nil {
+		log.Get(r.Context()).Debug("Failed to extract opentracing context on an incoming HTTP request.")
+	}
+	span, spanCtx := opentracing.StartSpanFromContext(r.Context(), utils.FetchRoutePathTemplate(r), ext.RPCServerOption(wireContext))
+	span = span.SetTag("http.method", r.Method)
+	span = span.SetTag("http.url", r.URL.String())
+	span = span.SetTag("peer.address", r.RemoteAddr)
```
briancharous: same comment as the logging one above

log/middleware.go
benjigoldberg: good catch, copy-paste error from old stuff

http/utils/utils.go
benjigoldberg: yeah I agree, struggled with what to name this. I'll thing about it more.

log/middleware.go
benjigoldberg: "RemoteAddr allows HTTP servers and other software to record the network address that sent the request, usually for logging." from official docs

Its a good question, I'm not sure if this gets mangled by ingress or not. We can check the logs.

benjigoldberg: Yeah, its the ingress. Rm'd since this is worthless.

PR #28, opened 2019-04-22
one-liner for a prod-compatible service

benjigoldberg:
@briancharous what do you think of this approach? I think we define one for http, and someday one for grpc. A user could take the two cobra commands and chain them together, as well.

### DIFF
--- a/service/http.go
+++ b/service/http.go
@@ -28,6 +28,7 @@ import (
 	"go.uber.org/zap/zapcore"
 )
 
+// HTTPConfig defines service level configuration for HTTP servers
 type HTTPConfig struct {
 	Name             string                                                             // Name of the application server
 	Version          string                                                             // Semantic Version of this Application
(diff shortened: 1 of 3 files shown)


service/http.go
```diff
@@ -0,0 +1,88 @@
...
+		Fields: map[string]interface{}{
+			"version": hc.Version,
+			"git_sha": hc.GitSHA,
+		},
+		Cores: []zapcore.Core{&tools.SentryCore{}},
+	}
+	cmd := &cobra.Command{
+		Use:              hc.Name,
+		Short:            "Starts and runs an HTTP Server",
+		Long:             `Starts and runs an HTTP Server`,
```
briancharous: Change to standard quotes

service/http.go
```diff
@@ -0,0 +1,88 @@
...
+		Version:          fmt.Sprintf("%s (%s)", hc.Version, hc.GitSHA),
+		PersistentPreRun: tools.CobraBindEnvironmentVariables(hc.Name),
+		RunE: func(cmd *cobra.Command, args []string) error {
```
briancharous: If this can't return an error just use `Run` instead of `RunE`

benjigoldberg:
also codecov is full of it, coverage on kafka did not go down lol

PR #29, opened 2019-04-23
Feature/go mod

benjigoldberg:
@briancharous here it is -- the move to go modules and sectioning tools into a series of smaller packages. Its a big one. Going to be work downstream to make this happen.

Also of note -- I deleted redundant http and log stuff from this as well.


benjigoldberg:
/bump minor -- because we're still pre-1.0

PR #30, opened 2019-04-23
service CLI improvements

benjigoldberg:
@briancharous I realized I glossed over some of the CLI config for the generic http service. This just adds that back and makes it a bit more common.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
 .PHONY: default_target all build test coverage clean lint
 
+VERSION ?= $(shell git describe --abbrev=0 --tags | sed 's/v//g')
+GIT_SHA ?= $(shell git rev-parse HEAD)
 LINTER_INSTALLED := $(shell sh -c 'which golangci-lint')
 
 default_target: all
@@ -10,7 +12,7 @@ tidy:
 	go mod tidy
 
 build: tidy
-	go build -ldflags="-X main.version=${VERSION} -X main.gitSha=${GIT_SHA}" examples/example_server.go
+	go build -ldflags="-X main.version=${VERSION} -X main.gitSHA=${GIT_SHA}" examples/example_server.go
 
 test: tidy
 	go test -race -v ./... -coverprofile=coverage.txt -covermode=atomic
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: default_target all build test coverage clean lint
+.PHONY: default_target all build test coverage lint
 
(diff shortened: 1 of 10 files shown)


service/cli.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/spf13/pflag"
+)
+
+// Config defines service level configuration for HTTP servers
+type Config struct {
+	Name        string               // Name of the application
+	Version     string               // Semantic Version of the application
+	GitSHA      string               // GitSHA of the application when compiled
+	Environment string               // Environment where the server is running
+	Package     string               // The name of this go package, eg `github.com/spothero/myservice`
```
briancharous: This is used for sentry only right? Not sure if this is possible but can we nix it here and use reflection to get this value elsewhere?

service/cli.go
```diff
@@ -0,0 +1,66 @@
...
+	Registry    *prometheus.Registry // The Prometheus Registry to use. If nil, the global registry is used by default.
+}
+
+// RegisterFlags registers Service flags with pflags
+func (c *Config) RegisterFlags(flags *pflag.FlagSet) {
+	flags.StringVarP(&c.Name, "name", "n", c.Name, "Name of the application")
+	flags.StringVarP(&c.Version, "version", "v", c.Version, "Semantic Version of the application")
+	flags.StringVarP(&c.GitSHA, "git-sha", "g", c.GitSHA, "Git SHA of the application")
```
briancharous: Adding Git SHA and Version as cli options is weird - they should be baked in with linker flags.

benjigoldberg: was trying to enforce this, but I agree with the argument.

Makefile
```diff
@@ -10,7 +12,7 @@ tidy:
 	go mod tidy
 
 build: tidy
-	go build -ldflags="-X main.version=${VERSION} -X main.gitSha=${GIT_SHA}" examples/example_server.go
+	go build -ldflags="-X main.version=${VERSION} -X main.gitSHA=${GIT_SHA}" examples/example_server.go
```
briancharous: remove the clean target since there's no vendoring

sentry/sentry.go
```diff
@@ -42,7 +42,12 @@ func (c Config) InitializeRaven() error {
 	}
 	raven.SetEnvironment(c.Environment)
 	raven.SetRelease(c.AppVersion)
-	appPackage = c.AppPackage
+
+	buildInfo, ok := debug.ReadBuildInfo()
```
briancharous: fancy

PR #39, opened 2019-04-29
Feature/pg init

benjigoldberg:
b/c spotbot closed the [other one due to a bug :(](https://github.com/spothero/tools/pull/38). WIP


benjigoldberg:
@briancharous @wilsoniya alright my dudes ready for review

sql/cli.go
```diff
@@ -0,0 +1,34 @@
...
+// limitations under the License.
+
+package sql
+
+import (
+	"github.com/spf13/pflag"
+)
+
+// RegisterFlags registers SQL flags with pflags
+func (pc *PostgresConfig) RegisterFlags(flags *pflag.FlagSet) {
```
briancharous: set some default values here, it's unlikely a bunch of this stuff needs to change, i.e. port, connection timeout, using ssl, etc. Also I think `lib/pq` picks up Postgres environment variables as well.

sql/metrics.go
```diff
@@ -0,0 +1,229 @@
...
+		maxIdleClosedConnections:     maxIdleClosedConnections,
+		maxLifetimeClosedConnections: maxLifetimeClosedConnections,
+		queryDuration:                queryDuration,
+		queryCount:                   queryCount,
+	}
+}
+
+// exportMetrics creates a goroutine which periodically scrapes the core database driver for
+// connection details, exporting those metrics for prometheus scraping
+func (m metrics) exportMetrics(db *sqlx.DB, frequency time.Duration) chan<- struct{} {
```
briancharous: Maybe a nitpick but I don't think I've seen empty struct used for signal channels like this. I've used `bool` elsewhere in this repo for similar purposes. 🤷‍♂

sql/middleware.go
```diff
@@ -0,0 +1,97 @@
...
+type ctxQueryNameKey int
+
+// ctxQueryNameValue is the value used to place the query name in context.Context
+const ctxQueryNameValue ctxQueryNameKey = iota
+
+// MiddlewareEnd is called after the SQL query has completed
+type MiddlewareEnd func(ctx context.Context, queryName, query string, queryErr error, args ...interface{}) (context.Context, error)
+
+// MiddlewareStart is called before the SQL query has started
+type MiddlewareStart func(ctx context.Context, queryName, query string, args ...interface{}) (context.Context, MiddlewareEnd, error)
```
briancharous: ha the github syntax highliter broke on this, I wonder why

sql/postgres.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/spothero/tools/log"
+	"go.uber.org/zap"
+)
+
+// pgDriverName is the name used to register the wrapped postgres SQL Driver with sql interface
+var pgDriverName = "instrumentedPostgres"
+
+// pgWrapped keeps track of whether or not the postgres client has already been wrapped
+var pgWrapped = false
```
briancharous: Is it possible to make this a non-global? What if I want to connect to 2 different postgres databases? I think wrapping the connection would fail.

sql/postgres.go
```diff
@@ -0,0 +1,164 @@
...
+	SSL              bool          // If true, connect to the database with SSL
+	SSLCert          string        // Path to the SSL Certificate, if any
+	SSLKey           string        // Path to the SSL Key, if any
+	SSLRootCert      string        // Path to the SSL Root Certificate, if any
+	MetricsFrequency time.Duration // How often to export core database metrics
+	Middleware       Middleware    // List of SQL Middlewares to apply, if any
+}
+
+// NewPostgresConfig creates and return a default postgres configuration.
+func NewPostgresConfig(dbName string) PostgresConfig {
```
briancharous: Name this `NewDefaultPostgresConfig`

sql/postgres.go
```diff
@@ -0,0 +1,164 @@
...
+		Port:             5432,
+		Database:         dbName,
+		ConnectTimeout:   defaultTimeout,
+		MetricsFrequency: defaultMetricsFrequency,
+	}
+}
+
+// buildConnectionString transforms the PostgresConfig into a usable connection string for lib/pq.
+// If a missing or invalid field is provided, an error is returned.
+func (pc PostgresConfig) buildConnectionString() (string, error) {
```
briancharous: Surprised that this functionality doesn't exist in `lib/pq` or something

sql/cli.go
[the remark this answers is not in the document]
benjigoldberg: This should be called at `NewDefaultPostgresConfig` which supplies the defaults here.

sql/postgres.go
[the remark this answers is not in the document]
benjigoldberg: It **has** to break -- this registers the _driver_ -- doing that twice causes a panic. You can still create as many databases as you want.

sql/postgres.go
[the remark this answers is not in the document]
benjigoldberg: This is [how it works](https://godoc.org/github.com/lib/pq#hdr-Connection_String_Parameters), sadly.

sql/middleware.go
benjigoldberg: prob because its a type alias in a type alias. yo dawg I heard you like type alias

sql/postgres.go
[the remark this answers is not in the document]
briancharous: 👍

PR #40, opened 2019-05-01
Feature/pg tracing

benjigoldberg:


log/middleware.go
```diff
@@ -45,3 +47,30 @@ func Middleware(sr *writer.StatusRecorder, r *http.Request) (func(), *http.Reque
...
+		"attempting sql query",
+		zap.String("query_name", queryName),
+		zap.String("query", query),
+	)
+	mwEnd := func(ctx context.Context, queryName, query string, queryErr error, args ...interface{}) (context.Context, error) {
+		if queryErr != nil {
+			Get(ctx).Error(
+				"failed sql query",
+				zap.String("query_name", queryName),
+				zap.String("query", query),
```
briancharous: could this potentially include PII or is all that stuff stubbed out with `?` or whatever placeholder?

tracing/middleware.go
```diff
@@ -0,0 +1,102 @@
...
+)
+
+// HTTPMiddleware extracts the OpenTracing context on all incoming HTTP requests, if present. if
+// no trace ID is present in the headers, a trace is initiated.
+//
+// The following tags are placed on all incoming HTTP requests:
+// * http.method
+// * http.hostname
+// * http.port
+// * http.remote_address
```
briancharous: I thought this was taken out. Looks like it just stuck around in the comment

tracing/tracing.go
```diff
@@ -71,14 +68,10 @@ func (c Config) ConfigureTracer() io.Closer {
 	tracer, closer, err := jaegerConfig.NewTracer(
 		jaegercfg.Logger(jaegerzap.NewLogger(logger)))
 	if err != nil {
-		logger.Error("Couldn't initialize Jaeger tracer", zap.Error(err))
+		logger.Error("couldnt initialize jaeger tracer", zap.Error(err))
```
briancharous: Is the apostrophe gone because it messes up some quoted string somewhere?

log/middleware.go
benjigoldberg: thankfully no -- the `args` are what would hold the PII. Like you said, this would have `?` or in the case of PG, `$1`, `$2`, and so on. Great thing to ask about though.

tracing/tracing.go
benjigoldberg: nah, just keeping it simple to work with as far as parsing in sumo.

tracing/middleware.go
benjigoldberg: oops. I'll remove.

tracing/tracing.go
briancharous: super nitpicky but it bothers me, can you make it say "could not"

PR #51, opened 2019-06-21
make README work

dj80hd:
minor tweaks to make the build/run procedure in README work

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ all: lint test
 build: ## Builds application artifacts
 	go build -ldflags="-X main.version=${VERSION} -X main.gitSHA=${GIT_SHA}" examples/example_server.go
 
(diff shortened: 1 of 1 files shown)


benjigoldberg:
Thanks for the PR @dj80hd! Was the `make` or `make build` command not creating the `example_server` binary for you? If so, could you post what it did output?

dj80hd:
`make build` will build `./example_server`, but `make` alone will not.  

Possibly, since `make` defaults are `lint` and `test`, `test` should have a dependency on `build` ?

benjigoldberg:
@dj80hd I like that suggestion -- thanks for finding this. If you make that change I can go ahead and get it merged.

Thanks for the contribution!

dj80hd:
glad to help @benjigoldberg -- this is a great idea

PR #55, opened 2019-08-12
Refactor service startup so that configured flags are accessible downstream

briancharous:
I was running into problems trying to configure a server that needed the environment flag defined by `service.Config` since the `RegisterHandlers` function has to be provided when creating the server cobra command. But, the actual environment value is not actually available until after the service starts up and populates its config from the CLI and/or environment variables.

So, instead of providing the `RegisterHandlers` when configuring a `service.HTTPConfig`, a function to construct the an interface with a `RegisterHandlers` method is provided when creating the server command. Then, the `ServerCmd` method can instantiate the downstream object with its configuration already populated by the CLI and/or environment variables.

### DIFF
--- a/examples/example_server.go
+++ b/examples/example_server.go
@@ -49,13 +49,15 @@ func (h handler) helloWorld(w http.ResponseWriter, r *http.Request) {
 	defer span.Finish()
 
 	// NOTE: Here we write out some artisanal HTML. There are many other (better) ways to output data.
-	fmt.Fprintf(
-		w, `
+	fmt.Fprintf(w,
+		`
 <html>
 Hello World. What's the <a href='/best-language'>best language?</a></br>
 (I'm running in the %s environment)
 </html>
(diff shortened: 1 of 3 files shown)


examples/example_server.go
```diff
@@ -29,23 +29,33 @@ import (
...
 	// NOTE: This is an example of an opentracing span
 	span, _ := opentracing.StartSpanFromContext(r.Context(), "example-hello-world")
 	span = span.SetTag("Key", "Value")
 	defer span.Finish()
 
 	// NOTE: Here we write out some artisanal HTML. There are many other (better) ways to output data.
-	fmt.Fprintf(w, "<html>Hello World. What's the <a href='/best-language'>best language?</a></html>")
+	fmt.Fprintf(
+		w, `
+<html>
```
benjigoldberg: how dare you reformat this beautiful html code!!!!!

PR #58, opened 2019-08-20
Make Kafka logging optional

briancharous:
This PR makes logging optional because if the user of the Kafka utils wants to handle the errors and log themselves, there are a lot of situations in which this library logging on its own results in duplicate logs.

### DIFF
--- a/kafka/kafka.go
+++ b/kafka/kafka.go
@@ -74,8 +74,8 @@ type Client struct {
 
 // ClientIface is an interface for creating consumers and producers
 type ClientIface interface {
-	NewConsumer() (ConsumerIface, error)
-	NewProducer(returnMessages bool) (ProducerIface, error)
+	NewConsumer(logger *zap.Logger) (ConsumerIface, error)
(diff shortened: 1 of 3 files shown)


kafka/kafka.go
```diff
@@ -385,7 +396,7 @@ func (c Consumer) ConsumeTopic(
 	catchupWg *sync.WaitGroup,
 	exitAfterCaughtUp bool,
 ) error {
-	log.Get(ctx).Info("Starting Kafka consumer", zap.String("topic", topic))
```
greg-mclain: Do we care that we're losing the context specific logging here? For kafka logs it may not matter that much, I'm just thinking out loud here...

briancharous: hmm yeah maybe it should log things that the consumer is doing but not successes and/or errors if the user wants to manage those themselves?

briancharous: Actually no, scratch the above comment, I think that might be too confusing of behavior. But no, I don't think there should be an issue losing context unless you wanted to provide different log fields when calling `ConsumeTopic` than were provided when calling `NewConsumer`.

PR #59, opened 2019-08-20
Allow users to store custom logger in ctx

briancharous:
Changes the `NewContext` function in the `log` package to accept a logger instead of fields to allow more customization. For example, one might want to pass a sub-logger or a logger with additional cores or options along inside the context instead of just a logger with pre-determined fields.

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -115,13 +115,16 @@ func (c Config) InitializeLogger() error {
 	return nil
 }
 
-// NewContext creates and returns a new context with a wrapped logger. If fields are specified,
-// all future invocations of the context logger will include those fields as well. This concept is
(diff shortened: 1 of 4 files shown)


briancharous:
Made one more update: the HTTP middleware now namespaces itself. This makes logs from the middleware look like:
```
2019-08-20T13:12:16.007-0500    INFO    log/middleware.go:51    request received        {"git_sha": "ot-set", "version": "not-set", "http": {"method": "GET", "path": "/health", "query_string": "", "correlation_id": "196e608adf227c7e"}}
2019-08-20T13:12:16.007-0500    INFO    log/middleware.go:54    returning response      {"git_sha": "ot-set", "version": "not-set", "http": {"response_code": 200}}
```

PR #60, opened 2019-08-21
Kafka instantiation refactor and package reorg

briancharous:
In this PR I moved a few things around to make it easier to pick and choose specific functionality of `kafka` package.

* Client vs Consumer vs Producer configuration is broken out of one giant config into individual config objects
* `NewConsumer` exposes the underlying Sarama client
* Metrics are still included in the wrapped consumer and producers, but can be used manually if interacting with Sarama directly
* Broker metrics exists separately from the producer and consumer metrics
* Code is reorganized from one giant file into files for consumer, producer, and client
* Errors are returned to the caller instead of panic-ing if something fails to create.


kafka/consumer.go
```diff
@@ -0,0 +1,387 @@
...
+
+// PartitionOffsets is a mapping of partition ID to an offset to which a consumer read on that partition
+type PartitionOffsets map[int32]int64
+
+// ConsumeTopic consumes a particular Kafka topic from startOffset to endOffset or
+// from startOffset to forever
+//
+// This function will create consumers for all partitions in a topic and read
+// from the given offset on each partition to the latest offset when the consumer was started, then notify the caller
+// via catchupWg. If exitAfterCaughtUp is true, the consumer will exit after it reads message at the latest offset
```
greg-mclain: Explanation of `exitAfterCaughtUp` is in this paragraph twice

PR #62, opened 2019-09-02
Add JOSE JWT parser and jwt-cli

benjigoldberg:
This begins the process of adding JWT and JSON Object Signing/Encryption (JOSE) support in tools.

This PR integrates functionality for Parsing JWT and configuring the JOSE client to point to a JWKS URL and extract the JWKS keyset. 

Incoming JWT tokens are decoded and validated against the JWK public key. Additionally, the issuer and expiration times are validated.

I didn't have time to get to building middleware to inspect `Authorization` headers, but I'll get to that next week. Basically my idea is to inspect the Auth headers in the middleware, call `ParseJWT` with `CognitoClaim` and then write an interface for each claim type (in this case, just cognito) that unpacks some relevant JWT user data (ie ClientID) into the HTTP context.


jose/jwt.go
```diff
@@ -0,0 +1,98 @@
...
+		return JOSE{}, err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		return JOSE{}, fmt.Errorf("received non-200 response from jwks url `%v`", resp.Status)
+	}
+
+	// Decode the response body into a JSONWebKeySet
+	jwks := &jose.JSONWebKeySet{}
+	err = json.NewDecoder(resp.Body).Decode(&jwks)
```
briancharous: I don't think you need a double pointer here

jose/jwt.go
```diff
@@ -0,0 +1,98 @@
...
+// NewJOSE creates and returns a JOSE client for use.
+func (c Config) NewJOSE() (JOSE, error) {
+	if len(c.JSONWebKeySetURL) == 0 {
+		return JOSE{}, fmt.Errorf("no jwks url specified and jwt signature verification enabled")
+	}
+
+	// Fetch JSON Web Key Sets from the specified URL
+	resp, err := http.Get(c.JSONWebKeySetURL)
+	if err != nil {
+		return JOSE{}, err
```
briancharous: Up to you, but I've started wrapping errors with `xerrors` since you can keep the originating error but provide more context. It isn't amazing but especially with Go 1.13 in RC, seems to be the way forward.

jose/jwt.go
```diff
@@ -0,0 +1,98 @@
...
+		jwks:        jwks,
+		validIssuer: c.ValidIssuer,
+	}, nil
+}
+
+// ParseJWT accepts a string containing a JWT token and attempts to parse and validate the token.
+// If you wish to inspect other components of the payload, you may supply one or more claims
+// structs which will be populated if the JWT is valid. Claims must be structs with json fields
+// that match the keys in the payload field, or a map[string]interface{}. Use of
+// map]string]interface{} is strongly discouraged.
```
briancharous: reversed `]`

greg-mclain:
The code looks good to me. Please just add a quick update to the README to indicate that this library now has support for JWT/JOSE

jose/jwt.go
benjigoldberg: yep, nice catch

jose/jwt.go
```diff
@@ -0,0 +1,99 @@
...
+package jose
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"time"
+
+	"golang.org/x/xerrors"
+	"gopkg.in/square/go-jose.v2"
```
wilsoniya: Is this used?

jose/jwt.go
```diff
@@ -0,0 +1,99 @@
...
+	TokenUse string `json:"token_use"`
+	Scope    string `json:"scope"`
+	ClientID string `json:"client_id"`
+	Version  int    `json:"version"`
+}
+
+// NewJOSE creates and returns a JOSE client for use.
+func (c Config) NewJOSE() (JOSE, error) {
+	if len(c.JSONWebKeySetURL) == 0 {
+		return JOSE{}, fmt.Errorf("no jwks url specified and jwt signature verification enabled")
```
wilsoniya: nit: capitalization of initialisms, i.e., jwks, url, jwt

Also: should `xerrors` be used here for consistency?

jose/jwt.go
```diff
@@ -0,0 +1,99 @@
...
+		validIssuer: c.ValidIssuer,
+	}, nil
+}
+
+// ParseJWT accepts a string containing a JWT token and attempts to parse and validate the token.
+// If you wish to inspect other components of the payload, you may supply one or more claims
+// structs which will be populated if the JWT is valid. Claims must be structs with json fields
+// that match the keys in the payload field, or a map[string]interface{}. Use of
+// map[string]interface{} is strongly discouraged.
+func (j JOSE) ParseJWT(input string, claims ...interface{}) error {
```
wilsoniya: This does more than parses a string to a JWT, it also validates. 
* at minimum rename the function to, e.g., `ParseJWTAndValidate`
* or break this into separate functions, i.e., one for parsing, another for validating.

jose/jwt.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	return JOSE{
+		jwks:        jwks,
+		validIssuer: c.ValidIssuer,
+	}, nil
+}
+
+// ParseJWT accepts a string containing a JWT token and attempts to parse and validate the token.
+// If you wish to inspect other components of the payload, you may supply one or more claims
```
wilsoniya: It's pretty unclear to me how the `claims` elements are used. They seem to be modified in-place? That seems weird, but maybe I'll get a better picture of what's going on here when i look at the tests. 

EDIT: It seems like there's no test coverage of `ParseJWT` with claims structures. That's something we should add for its own sake, plus it would serve as an example.

jose/jwt.go
benjigoldberg: @wilsoniya errors and logs are all lower-case

jose/jwt.go
benjigoldberg: The test coverage is actually a substantial ask @wilsoniya, we'd have to generate a full RSA keychain, generate the JWT, sign it, and then encode that into the file -- that's why I've chosen not to add coverage for it.

Makefile
```diff
@@ -1,13 +1,18 @@
-.PHONY: default_target all build test coverage lint help
+.PHONY: default_target all build test coverage lint help example-server jwt-cli
```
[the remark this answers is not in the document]
benjigoldberg: thanks for the suggestion -- I'm actually aware of this but don't like how it pollutes the file, personally. But, given that this list is growing quite quickly, I will update to what you've suggested.

jose/jwt.go
benjigoldberg: Yes, its used for `jose.JSONWebKeySet` on line 37

jose/jwt.go
benjigoldberg: renamed `ParseValidateJWT`

PR #63, opened 2019-09-04
Add RequiredAcks config to producer

greg-mclain:


kafka/producer.go
```diff
@@ -47,16 +47,30 @@ type ProducerIface interface {
 	Errors() chan *sarama.ProducerError
 }
 
+
+// Constant values that represent the required acks setting for produced messages. These map to
+// the sarama.RequiredAcks constants
+const (
```
benjigoldberg: Why duplicate when we could just use what Sarama has already defined?

https://github.com/Shopify/sarama/blob/master/produce_request.go#L10

kafka/producer.go
```diff
@@ -47,16 +47,30 @@ type ProducerIface interface {
...
+	Local
+	// None waits for no acknowledgements.
+	None
+)
+
 // ProducerConfig contains producer-specific configuration information
 type ProducerConfig struct {
 	ProducerCompressionCodec string
 	ProducerCompressionLevel int
+	ProducerRequiredAcks int
```
benjigoldberg: we should make this `sarama.RequiredAcks` 
See: 

1. https://github.com/Shopify/sarama/blob/master/produce_request.go#L10
2. https://godoc.org/github.com/Shopify/sarama#RequiredAcks

kafka/producer.go
greg-mclain: good point. I originally made the config param an int because that is what is supported by cobra when parsing config, not really thinking that I could use a local var to capture the command line param and then store it in the producer config. It's all fixed up now.

PR #64, opened 2019-09-11
Feature/jwt http middleware

benjigoldberg:
@wilsoniya @greg-mclain adds JWT parsing middleware. This will place any claims in the context if any are found.


jose/jwt.go
```diff
@@ -30,24 +31,34 @@ type Config struct {
 	ValidIssuer      string // URL of the JWT Issuer for this environment
 }
 
-// JOSE contains configuration for handling JWTs, JWKS, and other JOSE specifications
-type JOSE struct {
-	validIssuer string
-	jwks        *jose.JSONWebKeySet
+// ClaimGenerator defines an interface which creates a JWT Claim
+type ClaimGenerator interface {
+	New() Claim
```
wilsoniya: Interface members should get documentation.

jose/jwt.go
```diff
@@ -30,24 +31,34 @@ type Config struct {
...
+}
+
+// Claim defines an interface for common JWT claim functionality, such as registering claims to
+// contexts.
+type Claim interface {
+	WithContext(c context.Context) context.Context
```
wilsoniya: Interface members should get documentation.

wilsoniya: This applies to any other new interface methods.

jose/jwt.go
```diff
@@ -73,6 +84,16 @@ func (c Config) NewJOSE() (JOSE, error) {
 	}, nil
 }
 
+// GetClaims generates a claim for each registered claim generator and returns the newly generated
```
wilsoniya: `JOSE.claimGenerators` is a non-exported member, so mentioning it the docs here is leaking implementation details.

jose/jwt.go
```diff
@@ -30,24 +31,34 @@ type Config struct {
...
-	Version  int    `json:"version"`
+// JOSEHandler defines an interface for interfacing with JOSE and JWT functionality
+type JOSEHandler interface {
+	GetClaims() []Claim
+	ParseValidateJWT(input string, claims ...interface{}) error
+}
+
+// JOSE contains configuration for handling JWTs, JWKS, and other JOSE specifications
+type JOSE struct {
+	claimGenerators []ClaimGenerator
```
wilsoniya: Is this attribute ever being written to? I only see reads in this file.

jose/middleware.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+// authHeader defines the name of the header containing the JWT authorization data
+const authHeader = "Authorization"
+
+// bearerPrefix defines the standard expected form for OIDC JWT Tokens in Authorization headers.
+// Eg `Authorization: Bearer <JWT>`
+const bearerPrefix string = "Bearer "
+
+// HTTPMiddleware extracts the Authorization header, if present, on all incoming HTTP requests.
```
wilsoniya: `HTTPMiddleware` -> `GetHTTPMiddleware`

jose/middleware.go
```diff
@@ -0,0 +1,66 @@
...
+	return func(sr *writer.StatusRecorder, r *http.Request) (func(), *http.Request) {
+		logger := log.Get(r.Context())
+		authHeader := r.Header.Get(authHeader)
+		if len(authHeader) == 0 {
+			logger.Debug("no authorization header found")
+			return func() {}, r
+		}
+
+		if !strings.HasPrefix(authHeader, bearerPrefix) {
+			logger.Error("authorization header did not include bearer prefix")
```
wilsoniya: nit: this seems more like a warning or even info. A client sending a malformed auth header doesn't really indicate an error condition within the program.

jose/middleware.go
```diff
@@ -0,0 +1,66 @@
...
+			return func() {}, r
+		}
+
+		claims := jh.GetClaims()
+		err := jh.ParseValidateJWT(strings.TrimPrefix(authHeader, bearerPrefix), claims)
```
wilsoniya: I'd break `strings.TrimPrefix(authHeader, bearerPrefix)` into its own variable for a little bit more implicit documentation and call it e.g. `bearerToken`.

jose/cognito.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"context"
+)
+
+// cogntioCtxKey is the type used to uniquely place the cognito claim in the context
+type cognitoCtxKey int
+
+// CognitoClaimKey is the value used to uniquely place the cognito claim within the context
+const CognitoClaimKey cognitoCtxKey = iota
```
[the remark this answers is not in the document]
benjigoldberg: yeah, good call

jose/middleware.go
benjigoldberg: You're totally right -- this is a client error, not ours. I'm going to actually change this to Debug. Thanks for catching this oversight.

jose/jwt.go
benjigoldberg: another good catch 👏

jose/cognito.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// New satisfies the ClaimGenerator interface, returning an empty claim for use with JOSE parsing
+// and validation.
+func (cg CognitoGenerator) New() Claim {
+	return &CognitoClaim{}
+}
+
+// WithContext registers a claim to a given context
+func (cc CognitoClaim) WithContext(ctx context.Context) context.Context {
```
greg-mclain: It's minor, but how do you feel about renaming this since it's not consistent with how it's used in other libraries? (e.g. https://tip.golang.org/pkg/net/http/#Request.WithContext). I can't really think of a good name though. Maybe `CognitoClaimContext` or something like that. It's up to you...

jose/jose_mock.go
```diff
@@ -0,0 +1,64 @@
...
+	for i, generator := range mh.claimGenerators {
+		claims[i] = generator.New()
+	}
+	return claims
+}
+
+// ParseValidateJWT mocks the ParseValidateJWT function
+func (mh *MockHandler) ParseValidateJWT(input string, claims ...interface{}) error {
+	return mh.Called(input, claims).Error(0)
+}
```
greg-mclain: Out of curiosity, did you try out that mock code generator for this?

jose/middleware.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/spothero/tools/log"
+	"go.uber.org/zap"
+)
+
+// authHeader defines the name of the header containing the JWT authorization data
+const authHeader = "Authorization"
+
+// bearerPrefix defines the standard expected form for OIDC JWT Tokens in Authorization headers.
+// Eg `Authorization: Bearer <JWT>`
+const bearerPrefix string = "Bearer "
```
greg-mclain: nit: declare these consts consistently. One has the `string` type declared and the other doesn't.

jose/jose_mock.go
benjigoldberg: nope, found it after the fact

jose/cognito.go
benjigoldberg: I think its a good call to rename this. I'm going to rename `NewContext` because we're returning a new context.

PR #72, opened 2019-09-27
Sentry middleware

briancharous:
This puts the Sentry recovery middleware back in place and amends the logger so that error logs that get sent to Sentry are much richer since they have request metadata. Sentry also captures the opentracing trace ID if possible and adds it to the downstream Sentry scope.

### DIFF
--- a/sentry/middleware.go
+++ b/sentry/middleware.go
@@ -8,6 +8,10 @@ import (
 	"github.com/spothero/tools/log"
 )
 
+// This middleware is a wrapper around the sentry-go library's middleware that attaches a the Sentry hub that
+// from the request context to the logger. That way, if the logger ever writes an error log, instead of just sending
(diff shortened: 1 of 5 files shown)


sentry/middleware.go
```diff
@@ -0,0 +1,35 @@
...
+	})
+	return sentryHandler.HandleFunc(func(w http.ResponseWriter, r *http.Request) {
+		hub := sentry.GetHubFromContext(r.Context())
+		hub.ConfigureScope(func(scope *sentry.Scope) {
+			span := opentracing.SpanFromContext(r.Context())
+			if sc, ok := span.Context().(jaeger.SpanContext); ok {
+				scope.SetTag("trace_id", sc.TraceID().String())
+			}
+		})
+		ctx := log.NewContext(r.Context(), log.Get(r.Context()).With(Hub(hub)))
```
benjigoldberg: /giphy exploding head

PR #79, opened 2019-10-01
GetCorrelationID() result now non-nillable for safety/ergonomics

wilsoniya:
This PR adjusts the implicit contract of `GetCorrelationID()` such that the function now avoids returning a nillable `string` type.

### DIFF
--- a/tracing/middleware.go
+++ b/tracing/middleware.go
@@ -29,11 +29,11 @@ import (
 	"go.uber.org/zap"
 )
 
-// correlationIDCtxKey is the key into the Context of the wrapped HTTP request
+// CorrelationIDCtxKey is the key into the Context of the wrapped HTTP request
 // which maps to the correlation id of the request. This correlation ID can be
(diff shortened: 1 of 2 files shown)


tracing/middleware.go
```diff
@@ -29,11 +29,11 @@ import (
 	"go.uber.org/zap"
 )
 
-// correlationIDCtxKey is the key into the Context of the wrapped HTTP request
+// CorrelationIDCtxKey is the key into the Context of the wrapped HTTP request
 // which maps to the correlation id of the request. This correlation ID can be
 // conveyed to external clients in order to correlate external systems with
 // SpotHero tracing and logging.
-const correlationIDCtxKey = "correlationIDCtxKey"
+const CorrelationIDCtxKey = "CorrelationIDCtxKey"
```
wilsoniya: This is now public because it is useful for testing and for for additional flexibility.

PR #80, opened 2019-10-07
Http query parsers

wilsoniya:
This PR introduces reusable query parameter parsers, specifically:

- ParseCoordinates
- ParseTime

@briancharous @greg-mclain just CCing you in on this PR as tools users

### DIFF
--- a/http/parsers.go
+++ b/http/parsers.go
@@ -26,6 +26,13 @@ type Coordinates struct {
 	longitude float64
 }
 
+// ParseCoordinates reads and parses from from the query parameters to the
+// supplied the supplied request latitude and logitude corresponding to
+// latFieldName and lonFieldName, respectively. An error is returned if only
+// one of the named fields is present, if either value cannot be parsed as a
+// float, or if either value is out of range for decimal latitude and
+// longitude. The returned struct reference is nil if none of latFieldName or
+// lonFieldName are present in the query parameters to the given request.
 func ParseCoordinates(r *http.Request, latFieldName, lonFieldName string) (*Coordinates, error) {
 	r.ParseForm()
 	latStr := r.Form.Get(latFieldName)
@@ -70,6 +77,11 @@ func ParseCoordinates(r *http.Request, latFieldName, lonFieldName string) (*Coor
 	}
 }
 
+// ParseTime reads and parses from the query parameters to the supplied request
+// a Time value corresponding to fieldName. An error is returned if a Time
+// could not be parsed from the given field. The zero-valued Time is returned
+// if the given field is not present in the query parameters to the supplied
+// request.
 func ParseTime(r *http.Request, fieldName string) (time.Time, error) {
 	r.ParseForm()
 	fieldStr := r.Form.Get(fieldName)
--- a/http/parsers.go
+++ b/http/parsers.go
@@ -21,6 +21,8 @@ import (
 	"time"
 )
 
+// Coordinates is an encapsulation of geospatial coordinates in decimal
+// degrees.
 type Coordinates struct {
 	latitude  float64
 	longitude float64
--- a/http/parsers.go
+++ b/http/parsers.go
@@ -15,10 +15,11 @@
 package http
 
 import (
-	"fmt"
 	"net/http"
 	"strconv"
 	"time"
+
+	"golang.org/x/xerrors"
 )
 
 // Coordinates is an encapsulation of geospatial coordinates in decimal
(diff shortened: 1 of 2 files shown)


http/parsers.go
```diff
@@ -0,0 +1,102 @@
...
+	latStr := r.Form.Get(latFieldName)
+	lonStr := r.Form.Get(lonFieldName)
+	latPresent, lonPresent := false, false
+	var lat, lon float64
+
+	if latStr != "" {
+		// case: some value supplied for latitude
+		_lat, errLat := strconv.ParseFloat(latStr, 64)
+		if errLat != nil {
+			return nil, fmt.Errorf("unable to parse %v", latFieldName)
```
benjigoldberg: We should likely wrap the error instead with [`xerrors`: `xerrors.Errorf`](https://godoc.org/golang.org/x/xerrors)

This applies for a number of errors below, I'll just make the comment once

http/parsers.go
```diff
@@ -0,0 +1,102 @@
...
+
+	if lat < -90 || lat > 90 {
+		return nil, fmt.Errorf("%v must be in [-90, 90]", latFieldName)
+	}
+
+	if lon < -180 || lon > 180 {
+		return nil, fmt.Errorf("%v must be in [-180, 180]", lonFieldName)
+	}
+
+	if latPresent && lonPresent {
```
benjigoldberg: This is a minor idiomatic suggestion here, as suggested in [Effective Go](https://golang.org/doc/effective_go.html#control-structures), but we tend to prefer when conditionals are used to encapsulate degenerate cases only. Eg, the function should end in the "happy-path" return value if possible, eg:

```
if latPresent != lonPresent {
    return nil, fmt.Errorf("both %v and %v must be provided", latFieldName, lonFieldName)
} else if !latPresent && !lonPresent {
    return nil, nil
}
return &Coordinates{lat, lon}, nil
```
Additionally, is it not an error if no lat and lon is found?

http/parsers.go
```diff
@@ -0,0 +1,102 @@
...
+
+// ParseCoordinates reads and parses from the query parameters to the supplied
+// request latitude and logitude corresponding to latFieldName and
+// lonFieldName, respectively. An error is returned if only one of the named
+// fields is present, if either value cannot be parsed as a float, or if either
+// value is out of range for decimal latitude and longitude. The returned
+// struct reference is nil if none of latFieldName or lonFieldName are present
+// in the query parameters to the given request.
+func ParseCoordinates(r *http.Request, latFieldName, lonFieldName string) (*Coordinates, error) {
+	r.ParseForm()
```
benjigoldberg: I wonder if parseform is a noop if its already been called on a request 🤔

http/parsers.go
```diff
@@ -0,0 +1,102 @@
...
+
+// ParseTime reads and parses from the query parameters to the supplied request
+// a Time value corresponding to fieldName. An error is returned if a Time
+// could not be parsed from the given field. The zero-valued Time is returned
+// if the given field is not present in the query parameters to the supplied
+// request.
+func ParseTime(r *http.Request, fieldName string) (time.Time, error) {
+	r.ParseForm()
+	fieldStr := r.Form.Get(fieldName)
+	if fieldStr != "" {
```
benjigoldberg: I would argue this should instead be
```
if fieldStr == "" {
    return time.Time{}, nil
}
... <rest of logic from if statement goes here>

http/parsers.go
wilsoniya: Not sure if it's a noop, but it is explicitly called out as idempotent.

wilsoniya: Sidebar, IMO it's really shitty API design forcing the consumer to call a method for its side effects in this way.

http/parsers.go
wilsoniya: Thanks for helping me w/ go idioms. 

> Additionally, is it not an error if no lat and lon is found?

I designed these parsers to *optionally* parse values from query strings, i.e., if the sought values are not present that's not considered an error. The rationale is that query params are often used to provide optional information e.g. for filters and sorting. IMO a missing value does not express an error, and furthermore, disambiguating an error indicating a missing value from any other error would make these APIs more difficult to use. 

Unfortunately Go does not make it easy to express "missing" values across interfaces. Where possible I opt to return zero-valued structs where that carries unambiguous meaning, elsewhere I'm opting to return nillable pointers (which is very unfortunate).

http/parsers.go
wilsoniya: Nice :+1:

http/parsers.go
benjigoldberg: ah, I see, ok.

http/parsers.go
benjigoldberg: yes, I agree

PR #82, opened 2019-12-04
Feature/grpc scaffold

benjigoldberg:
This PR adds a standardized GRPC Server and a default "SpotHero" service for our users to interface with to create GRPC servers quickly.

I haven't yet built the jaeger, logging, and metrics middleware. That will come in the next PR.


grpc/server.go
```diff
@@ -0,0 +1,114 @@
...
+			),
+		),
+		grpc.UnaryInterceptor(
+			grpc_middleware.ChainUnaryServer(
+				c.UnaryInterceptors...,
+			),
+		),
+	)
+	if c.ServerRegistration == nil {
+		return Server{}, fmt.Errorf("no server registration function provided")
```
briancharous: I feel like letting the call to `ServerRegistration` panic instead of returning an error is acceptable since `ServerRegistration` being `nil` is a usage error on the part of the user of this library and I can't really think of a scenario where this would be a recoverable situation. Getting rid of the error from this method would enable a nice 1-liner `err := c.NewServer().Run()` 🤷‍♂

grpc/server.go
```diff
@@ -0,0 +1,114 @@
...
+
+	"github.com/grpc-ecosystem/go-grpc-middleware"
+	"github.com/spothero/tools/log"
+	"go.uber.org/zap"
+	"google.golang.org/grpc"
+)
+
+// Config contains the configuration necessary for running a GRPC Server.
+type Config struct {
+	Name               string                         // Name of the HTTP Server
```
wilsoniya: s/HTTP/GRPC/

grpc/server.go
```diff
@@ -0,0 +1,114 @@
...
+	Address            string                         // Address on which the server will be accessible
+	Port               int                            // Port on which the server will be accessible
```
wilsoniya: technically should be `uint16`

service/grpc.go
```diff
@@ -0,0 +1,122 @@
...
+	"google.golang.org/grpc"
+)
+
+// GRPCConfig contains required configuration for starting a GRPC service
+type GRPCConfig struct {
+	Config
+	CancelSignals []os.Signal // OS Signals to be used to cancel running servers. Defaults to SIGINT/`os.Interrupt`.
+}
+
+// GRPCService implementers register GRPC APIs with the GRPC server
```
wilsoniya: sp: implementors

service/grpc.go
```diff
@@ -0,0 +1,122 @@
...
+// production-quality GRPC server. This method takes a function that instantiates the GRPCService
+// interface that passes through the GRPCConfig object to the constructor after all values are
+// populated from the CLI and/or environment variables so that values configured by this package
+// are accessible downstream.
+//
+// Note that this function returns the Default GRPC server for use at SpotHero. Consumers of the
+// tools libraries are free to define their own server entrypoints if desired. This function is
+// provided as a convenience function that should satisfy most use cases.
+//
+// Note that Version and GitSHA *must be specified* before calling this function.
```
wilsoniya: do we need to validate on this?

service/grpc.go
```diff
@@ -0,0 +1,122 @@
...
+	CancelSignals []os.Signal // OS Signals to be used to cancel running servers. Defaults to SIGINT/`os.Interrupt`.
+}
+
+// GRPCService implementers register GRPC APIs with the GRPC server
+type GRPCService interface {
+	ServerRegistration(*grpc.Server)
+}
+
+// ServerCmd creates and returns a Cobra and Viper command preconfigured to run a
+// production-quality GRPC server. This method takes a function that instantiates the GRPCService
```
wilsoniya: This sentence is confusing to me. How about: 

> This method takes a function, newService, that instantiates the GRPCService by consuming the GRPCConfig object after all values are populated from the CLI and/or environment variables so that values configured by this package are accessible by newService.

grpc/server.go
benjigoldberg: Updated

service/grpc.go
benjigoldberg: We already do here: https://github.com/spothero/tools/blob/feature/grpc-scaffold/service/grpc.go#L91

which calls this: https://github.com/spothero/tools/blob/feature/grpc-scaffold/service/cli.go#L43

service/grpc.go
benjigoldberg: I like it. I'm updating HTTP ServerCmd as well.

PR #84, opened 2019-12-09
Grpc tls

benjigoldberg:
This adds TLS for the HTTP server and GRPC. Previously we had configuration for TLS but never used it anywhere. This corrects that oversight. I also got an error from our tester indicating that there was a race condition sometimes in our log tests in terms of how `logger` was being modified. Given, that, I updated those tests to protect the base `logger` with a mutex.

Also: Note that I bump `go.mod` from `1.12` to `1.13`. We may want to consider removing the `x/errors` package from tools entirely.


grpc/server.go
```diff
@@ -92,14 +99,36 @@ func (c Config) NewServer() Server {
...
-	listener, err := net.Listen("tcp", s.listenAddress)
-	if err != nil {
-		log.Get(ctx).Error("error starting grpc server listener", zap.Error(err))
-		return err
+	var listener net.Listener
+	var err error
+	if s.tlsEnabled {
+		cert, err := tls.LoadX509KeyPair(s.tlsCrtPath, s.tlsKeyPath)
+		if err != nil {
+			log.Get(ctx).Error("failed to load tls x509 key pair", zap.Error(err))
```
briancharous: I'm not sure that logging is the appropriate thing to do in these error cases. Since you're returning the same error message with the originating error wrapped up, the most likely scenario is the caller is going to either log an error or die with the error message which will duplicate the log message.

http/server.go
```diff
@@ -131,9 +137,20 @@ func (s Server) Run() {
...
-			log.Get(ctx).Info("HTTP server shutdown", zap.Error(err))
+		var err error
+		if s.tlsEnabled {
+			log.Get(ctx).Info(fmt.Sprintf("HTTPS server started on %s", s.httpServer.Addr))
+			err = s.httpServer.ListenAndServeTLS(s.tlsCrtPath, s.tlsKeyPath)
+		} else {
+			log.Get(ctx).Info(fmt.Sprintf("HTTP server started on %s", s.httpServer.Addr))
+			err = s.httpServer.ListenAndServe()
+		}
+		fmt.Println(err)
```
briancharous: rm

http/server.go
```diff
@@ -131,9 +137,20 @@ func (s Server) Run() {
...
+		switch err {
+		case http.ErrServerClosed:
+			log.Get(ctx).Info("http server shutdown", zap.Error(err))
```
briancharous: Why log an error at an info log? The "error" here is the server is closed.

grpc/server.go
benjigoldberg: Agreed, I'll update

PR #86, opened 2019-12-09
Instrument the GRPC server and make request received log at Debug

benjigoldberg:
* Replaces the GRPC logger with our zap logger
* Makes `request received` log messages Debug, as we discussed, to cut down on excess logging


log/middleware.go
```diff
@@ -42,8 +42,7 @@ func HTTPMiddleware(next http.Handler) http.Handler {
 		method := zap.String("http_method", r.Method)
 		path := zap.String("path", writer.FetchRoutePathTemplate(r))
 		query := zap.String("query_string", r.URL.Query().Encode())
-		logger.Info("request received", method, path, query)
-		logger.Debug("request headers", zap.Reflect("Headers", r.Header))
```
benjigoldberg: Note that I removed header logging. This is a security concern w/ bearer tokens and others, imo.

PR #87, opened 2019-12-10
Grpc http merge

benjigoldberg:
This merges the grpc and http service into one service.


examples/web/server.go
```diff
@@ -34,14 +35,20 @@ type handler struct {
...
 // NOTE: The HTTP server automatically registers /health and /metrics -- Have a look in your
 // browser!
 func (h handler) RegisterHandlers(router *mux.Router) {
 	router.HandleFunc("/", h.helloWorld)
 	router.HandleFunc("/best-language", bestLanguage)
 }
 
+// ServerRegistration is a callback used to register GRPC endpoints to the default server.
+// The handler is empty since we are not registering any GRPC APIs in this example.
+func (h handler) ServerRegistration(server *grpc.Server) {
```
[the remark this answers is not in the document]
wilsoniya: +1

service/cli.go
```diff
@@ -14,22 +14,27 @@
...
-	CancelSignals []os.Signal           // OS Signals to be used to cancel running servers. Defaults to SIGINT/`os.Interrupt`.
+	Name          string                                                             // Name of the application
+	Environment   string                                                             // Environment where the server is running
+	Version       string                                                             // Semantic Version of the application
+	GitSHA        string                                                             // GitSHA of the application when compiled
+	Registry      prometheus.Registerer                                              // The Prometheus Registry to use. If nil, the global registry is used by default.
+	CancelSignals []os.Signal                                                        // OS Signals to be used to cancel running servers. Defaults to SIGINT/`os.Interrupt`.
+	PreStart      func(ctx context.Context, router *mux.Router, server *http.Server) // A function to be called before starting the HTTP web server
```
wilsoniya: here and on the next line you're mentioning HTTP, which seems overspecified given this will apply to both HTTP and GRPC services, yes?

service/service.go
```diff
@@ -51,41 +56,57 @@ type GRPCService interface {
 // provided as a convenience function that should satisfy most use cases.
 //
 // Note that Version and GitSHA *must be specified* before calling this function.
-func (gc GRPCConfig) ServerCmd(
+func (c Config) ServerCmd(
```
wilsoniya: FYI the doc comment on this func isn't golang style, i.e., doesn't start with the func name.

service/cli.go
benjigoldberg: This actually only applies to HTTP. I'll update the names of the functions to `PreStartHTTP` and `PostStartHTTP`

PR #96, opened 2020-01-02
Feature/http client

benjigoldberg:
Opening this up for feedback now. I'm not deeply satisfied with where this ended up. There are a couple of issues that need to be figured out.

First, the way that we create `http.Metrics` with `http.NewMetrics` will not work correctly with the default client. I need to step back and think of a more effective way of creating and sharing around that metrics bundle.

Second, I'm not happy at all with 

`func (mrt MiddlewareRoundTripper) RounTrip()...`

and the way it handles Middleware. I'm considering rewriting this to instead be a series of `http.RoundTripper` interfaces that nest eachother... But before I do that I want to get your perspective.

The reason I don't like this current approach is that it makes things like exponential backoff essentially impossible.

### DIFF
--- a/http/client.go
+++ b/http/client.go
@@ -3,6 +3,10 @@ package http
 import (
 	"fmt"
 	"net/http"
+
+	"github.com/spothero/tools/jose"
+	"github.com/spothero/tools/log"
+	"github.com/spothero/tools/tracing"
 )
 
 // ClientMiddleware defines an HTTP Client middleware function. The function is called prior to
@@ -15,6 +19,25 @@ type ClientMiddleware func(*http.Request) (*http.Request, func(*http.Response) e
 type MiddlewareRoundTripper struct {
 	RoundTripper http.RoundTripper
 	Middleware   []ClientMiddleware
+}
+
+// NewDefaultClient constructs the default HTTP Client with middleware. Providing an HTTP
+// RoundTripper is optional. If `nil` is received, the DefaultClient will be used.
+func NewDefaultClient(metrics Metrics, roundTripper http.RoundTripper) http.Client {
+	if roundTripper == nil {
+		roundTripper = http.DefaultTransport
+	}
+	return http.Client{
+		Transport: MiddlewareRoundTripper{
+			Middleware: []ClientMiddleware{
+				tracing.HTTPClientMiddleware,
+				log.HTTPClientMiddleware,
+				metrics.ClientMiddleware,
+				jose.HTTPClientMiddleware,
+			},
+			RoundTripper: roundTripper,
+		},
+	}
 }
 
 // RoundTrip completes the http request round trip and is responsible for invoking HTTP Client
@@ -49,13 +72,4 @@ func (mrt MiddlewareRoundTripper) RoundTrip(req *http.Request) (*http.Response, 
 		}
 	}
 	return resp, err
-}
-
-// BackoffClientMiddleware is middleware for use in HTTP Clients for automatically performing
-// exponential backoff and retries on failed HTTP requests
-func BackoffClientMiddleware(r *http.Request) (func(*http.Response) error, error) {
-	// TODO: Middleware!
-	return func(resp *http.Response) error {
-		return nil
-	}, nil
 }
--- a/http/client.go
+++ b/http/client.go
@@ -18,7 +18,16 @@ type ClientMiddleware func(*http.Request) (*http.Request, func(*http.Response) e
 // with middleware
 type MiddlewareRoundTripper struct {
 	RoundTripper http.RoundTripper
-	Middleware   []ClientMiddleware
+	// Middleware consist of a function which is called prior to the execution of a request. This
+	// function returns the potentially modified request, the post-response handler, and an error,
+	// if any. The response handler is invoked after the HTTP request has been made.
+	//
+	// Middleware are called in the order they are specified. In otherwords, the first item in the
+	// slice is the first middleware applied, and the last item in the slice is the last middleware
+	// applied. Each response handler is called in the reverse order of the middleware. Meaning,
+	// the last middleware called will be the first to have its response handler called, and
+	// likewise, the first middleware called will be the last to have its handler called.
+	Middleware []ClientMiddleware
 }
 
 // NewDefaultClient constructs the default HTTP Client with middleware. Providing an HTTP
@@ -45,17 +54,21 @@ func NewDefaultClient(metrics Metrics, roundTripper http.RoundTripper) http.Clie
 func (mrt MiddlewareRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 	// Ensure the RoundTripper was set on the MiddlewareRoundTripper
 	if mrt.RoundTripper == nil {
-		return nil, fmt.Errorf("no roundtripper provided to middleware round tripper")
+		panic("no roundtripper provided to middleware round tripper")
 	}
 
 	// Call all middleware
-	responseHandlers := make([]func(*http.Response) error, len(mrt.Middleware))
+	numMiddleware := len(mrt.Middleware)
+	responseHandlers := make([]func(*http.Response) error, numMiddleware)
 	for idx, middleware := range mrt.Middleware {
 		updatedReq, callback, err := middleware(req)
 		if err != nil {
 			return nil, fmt.Errorf("error invoking http client middleware: %w", err)
 		}
-		responseHandlers[idx] = callback
+		// Append handlers in reverse order so that nesting is reverse on response handling.
+		// Always call the last middleware's response handler first, the second to last
+		// middleware's response handler second, and so on.
+		responseHandlers[numMiddleware-idx-1] = callback
 		req = updatedReq
 	}
 
--- a/http/client.go
+++ b/http/client.go
@@ -8,30 +8,36 @@ import (
 // ClientMiddleware defines an HTTP Client middleware function. The function is called prior to
 // invoking the transport roundtrip, and the returned response function is called after the
 // response has been received from the client.
-type ClientMiddleware func(*http.Request) (func(*http.Response) error, error)
+type ClientMiddleware func(*http.Request) (*http.Request, func(*http.Response) error, error)
 
 // MiddlewareRoundTripper implements a proxied net/http RoundTripper so that http requests may be decorated
 // with middleware
 type MiddlewareRoundTripper struct {
-	roundTripper http.RoundTripper
+	RoundTripper http.RoundTripper
 	Middleware   []ClientMiddleware
 }
 
 // RoundTrip completes the http request round trip and is responsible for invoking HTTP Client
 // Middleware
 func (mrt MiddlewareRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	// Ensure the RoundTripper was set on the MiddlewareRoundTripper
+	if mrt.RoundTripper == nil {
+		return nil, fmt.Errorf("no roundtripper provided to middleware round tripper")
+	}
+
 	// Call all middleware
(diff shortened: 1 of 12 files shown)


log/middleware.go
```diff
@@ -37,26 +38,54 @@ import (
...
 			} else {
 				responseCodeField = zap.Skip()
 			}
-			logger.Info("returning response", responseCodeField)
+			logger.Info(
+				"http response returned",
+				responseCodeField,
+				method,
+				path,
+				zap.Duration("http.duration", time.Since(startTime)),
```
briancharous: should this just be `duration` since the log message already has "http"

log/middleware.go
```diff
@@ -37,26 +38,54 @@ import (
...
+	requestLogger := Get(r.Context())
+	logger := requestLogger.Named("http")
+	method := zap.String("http.method", r.Method)
+	path := zap.String("http.path", writer.FetchRoutePathTemplate(r))
+	query := zap.String("http.query", r.URL.Query().Encode())
+	logger.Debug("http request started", method, path, query)
+	return r, func(resp *http.Response) error {
+		logger.Info(
+			"http request completed",
+			zap.Int("http.status_code", resp.StatusCode),
```
briancharous: same question as above re `http.status_code` vs `status_code`

log/middleware.go
```diff
@@ -37,26 +38,54 @@ import (
...
+			method,
+			path,
+			zap.Duration("http.duration", time.Since(startTime)),
```
briancharous: this log message doesn't seem like it's going to be super meaningful if the url isn't in the message

http/metrics.go
```diff
@@ -100,3 +138,22 @@ func (m Metrics) Middleware(next http.Handler) http.Handler {
 		next.ServeHTTP(w, r)
 	})
 }
+
+// ClientMiddleware is middleware for use in HTTP Clients for capturing prometheus metrics
+func (m Metrics) ClientMiddleware(r *http.Request) (*http.Request, func(*http.Response) error, error) {
+	var receivedResp *http.Response
+	observerFunc := func(durationSec float64) {
+		labels := prometheus.Labels{
+			"path":        r.URL.Path,
```
briancharous: url should probably be in here

tracing/middleware.go
```diff
@@ -68,10 +68,27 @@ func HTTPMiddleware(next http.Handler) http.Handler {
 	})
 }
 
+// HTTPClientMiddleware is middleware for use in HTTP Clients
+func HTTPClientMiddleware(r *http.Request) (*http.Request, func(*http.Response) error, error) {
+	operationName := fmt.Sprintf("%s %s", r.Method, r.URL.String())
+	span, spanCtx := opentracing.StartSpanFromContext(r.Context(), operationName)
+	span = span.SetTag("http.method", r.Method)
+	span = span.SetTag("http.url", r.URL.String())
```
briancharous: add path here

briancharous:
Why would backoff and retries be impossible with this? Wouldn't retries and backoffs live in layer above the `http.RoundTripper` implementation?

http/client.go
```diff
@@ -0,0 +1,75 @@
...
+		updatedReq, callback, err := middleware(req)
+		if err != nil {
+			return nil, fmt.Errorf("error invoking http client middleware: %w", err)
+		}
+		responseHandlers[idx] = callback
+		req = updatedReq
+	}
+
+	// Make the request
+	resp, err := mrt.RoundTripper.RoundTrip(req)
```
theisenmark: What's stopping us from implementing the exponential backoff right here?

log/middleware.go
benjigoldberg: I picked this up from the open-source grpc logger and I prefer it. Sure, the logger name includes `grpc`, but its not particularly visible or useful. It also doesn't show up in the logs right now.

http/client.go
benjigoldberg: we could implement it here, sorry for the confusion to both you and @briancharous, I meant that it would not be easily implemented as a middleware

http/client.go
```diff
@@ -0,0 +1,75 @@
...
+			RoundTripper: roundTripper,
+		},
+	}
+}
+
+// RoundTrip completes the http request round trip and is responsible for invoking HTTP Client
+// Middleware
+func (mrt MiddlewareRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	// Ensure the RoundTripper was set on the MiddlewareRoundTripper
+	if mrt.RoundTripper == nil {
```
wilsoniya: This should never be able to happen and if it does indicates a serious but, so imo this should `panic`

http/client.go
```diff
@@ -0,0 +1,75 @@
...
+// MiddlewareRoundTripper implements a proxied net/http RoundTripper so that http requests may be decorated
+// with middleware
+type MiddlewareRoundTripper struct {
+	RoundTripper http.RoundTripper
+	Middleware   []ClientMiddleware
+}
+
+// NewDefaultClient constructs the default HTTP Client with middleware. Providing an HTTP
+// RoundTripper is optional. If `nil` is received, the DefaultClient will be used.
+func NewDefaultClient(metrics Metrics, roundTripper http.RoundTripper) http.Client {
```
wilsoniya: Why do we allow for customization of the `RoundTripper`? seems like request processing is exactly what the middlewares are for, no?

http/metrics.go
```diff
@@ -62,24 +79,56 @@ func NewMetrics(registry prometheus.Registerer, mustRegister bool) Metrics {
...
-			log.Get(context.Background()).Error("failed to register http histogram", zap.Error(err))
+		toRegister := map[string]prometheus.Collector{
+			"duration":       histogram,
+			"clientDuration": clientHistogram,
+			"counter":        counter,
+			"clientCounter":  clientCounter,
 		}
-		if err := registry.Register(counter); err != nil {
-			log.Get(context.Background()).Error("failed to register http counter", zap.Error(err))
+		for name, collector := range toRegister {
```
benjigoldberg: @briancharous I'm curious what you think of this. I'm updating `NewMetrics` so that it doesn't care if you reregister the same metrics multiple times. Essentially, if the registry already has the metric registered and `mustRegister` is `false`, we allow the construction to continue as a noop for that metric. That means we can instantiate `http.Metrics` on the same prometheus registry as many times as we want.

benjigoldberg:
@briancharous @theisenmark ready for another round

http/metrics.go
```diff
@@ -50,6 +52,20 @@ func NewMetrics(registry prometheus.Registerer, mustRegister bool) Metrics {
 			"status_code",
 		},
 	)
+	clientHistogram := prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name: "http_client_request_duration_seconds",
```
wilsoniya: is it safe to assume these metric names are somehow namespaced by the application using the tools lib?

jose/middleware.go
```diff
@@ -92,3 +93,12 @@ func GetHTTPMiddleware(jh JOSEHandler, authRequired bool) func(next http.Handler
 		})
 	}
 }
+
+// HTTPClientMiddleware is middleware for use in HTTP Clients which propagates the Authorization
+// headers
+func HTTPClientMiddleware(r *http.Request) (*http.Request, func(*http.Response) error, error) {
+	if jwtData, ok := r.Context().Value(JWTClaimKey).(string); ok {
+		r.Header.Set(authHeader, fmt.Sprintf("%s%s", bearerPrefix, jwtData))
```
wilsoniya: is the processing of the header value something that should be in a separate function?

log/middleware.go
```diff
@@ -37,26 +38,54 @@ import (
 // to show up in logs.
 func HTTPMiddleware(next http.Handler) http.Handler {
```
wilsoniya: idea: make an alias to this func called `HTTPServerMiddleware` and deprecate `HTTPMiddleware`

log/middleware.go
```diff
@@ -37,26 +38,54 @@ import (
...
+func HTTPClientMiddleware(r *http.Request) (*http.Request, func(*http.Response) error, error) {
+	startTime := time.Now()
+	requestLogger := Get(r.Context())
+	logger := requestLogger.Named("http")
+	method := zap.String("http.method", r.Method)
+	path := zap.String("http.path", writer.FetchRoutePathTemplate(r))
+	query := zap.String("http.query", r.URL.Query().Encode())
+	logger.Debug("http request started", method, path, query)
+	return r, func(resp *http.Response) error {
+		logger.Info(
```
wilsoniya: Probably would be useful to log `Content-Length`

http/metrics.go
```diff
@@ -27,8 +27,10 @@ import (
 
 // Metrics is a bundle of prometheus HTTP metrics recorders
 type Metrics struct {
-	counter  *prometheus.CounterVec
-	duration *prometheus.HistogramVec
+	counter        *prometheus.CounterVec
+	duration       *prometheus.HistogramVec
+	clientCounter  *prometheus.CounterVec
+	clientDuration *prometheus.HistogramVec
```
wilsoniya: might be useful to track `Content-Length` on a histogram

http/metrics.go
benjigoldberg: Yes, we automatically associate app name, container, etc to the metrics, in addition to adding labels that uniquely identify the requests

jose/middleware.go
benjigoldberg: why?

http/client.go
benjigoldberg: RoundTripper contains a lot of potential options we don't want to wrap. I want this roundtripper focused exclusively on middleware. [If you look at the options provided by `net/http Transport` (which implements `RoundTripper`)](https://golang.org/pkg/net/http/#Transport) there is extensive customization which users may want to provide that are unrelated to our middleware usage here.

http/client.go
```diff
@@ -0,0 +1,75 @@
...
+// ClientMiddleware defines an HTTP Client middleware function. The function is called prior to
+// invoking the transport roundtrip, and the returned response function is called after the
+// response has been received from the client.
+type ClientMiddleware func(*http.Request) (*http.Request, func(*http.Response) error, error)
+
+// MiddlewareRoundTripper implements a proxied net/http RoundTripper so that http requests may be decorated
+// with middleware
+type MiddlewareRoundTripper struct {
+	RoundTripper http.RoundTripper
+	Middleware   []ClientMiddleware
```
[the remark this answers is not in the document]
benjigoldberg: Great suggestion 👍

log/middleware.go
benjigoldberg: I like it -- I'm going to be more aggressive and just deprecate `HTTPMiddleware` because:

1. Basically all of our clients use our service interface so this is invisible
2. This is pre-1.0 software, so breaking changes can still happen between minor revs 😬

http/metrics.go
```diff
@@ -27,59 +28,114 @@ import (
...
+			Name: "http_client_requests_total",
+			Help: "Total number of HTTP Client Requests sent",
+		},
+		labels,
+	)
+	contentLength := prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name: "http_content_length_bytes",
+			Help: "HTTP Request content length histogram",
+			// Power of 2 bytes, starts at 1 byte and works up to 10MB
```
wilsoniya: technically 16MB, but close enuff

jose/middleware.go
wilsoniya: Only thought is that if we're doing this concatenation anywhere else we might want to extract it to a function to eliminate duplicate codepaths, but bc it's so simple i guess it's not necessary :man_shrugging:

http/metrics.go
benjigoldberg: good catch, fixed

PR #97, opened 2020-01-06
Feature/backoff

benjigoldberg:
Implements exponential backoff for the HTTP client on 5xx responses by default.

### DIFF
--- a/http/client.go
+++ b/http/client.go
@@ -3,10 +3,13 @@ package http
 import (
 	"fmt"
 	"net/http"
+	"time"
 
+	"github.com/cenkalti/backoff/v4"
 	"github.com/spothero/tools/jose"
 	"github.com/spothero/tools/log"
 	"github.com/spothero/tools/tracing"
+	"go.uber.org/zap"
 )
 
 // ClientMiddleware defines an HTTP Client middleware function. The function is called prior to
@@ -30,11 +33,43 @@ type MiddlewareRoundTripper struct {
 	Middleware []ClientMiddleware
 }
 
+// RetryRoundTripper wraps a roundtripper with retry logic
+type RetryRoundTripper struct {
+	RoundTripper         http.RoundTripper
+	RetriableStatusCodes []int
+	InitialInterval      time.Duration
+	RandomizationFactor  float64
+	Multiplier           float64
+	MaxInterval          time.Duration
+	MaxRetries           uint64
+}
+
 // NewDefaultClient constructs the default HTTP Client with middleware. Providing an HTTP
 // RoundTripper is optional. If `nil` is received, the DefaultClient will be used.
 func NewDefaultClient(metrics Metrics, roundTripper http.RoundTripper) http.Client {
 	if roundTripper == nil {
 		roundTripper = http.DefaultTransport
+	}
+	retryRoundTripper := RetryRoundTripper{
+		RoundTripper: roundTripper,
+		RetriableStatusCodes: []int{
+			http.StatusInternalServerError,
+			http.StatusNotImplemented,
+			http.StatusBadGateway,
+			http.StatusServiceUnavailable,
+			http.StatusGatewayTimeout,
+			http.StatusHTTPVersionNotSupported,
+			http.StatusVariantAlsoNegotiates,
+			http.StatusInsufficientStorage,
+			http.StatusLoopDetected,
+			http.StatusNotExtended,
+			http.StatusNetworkAuthenticationRequired,
+		},
+		InitialInterval:     100 * time.Millisecond,
+		Multiplier:          2,
+		MaxInterval:         30 * time.Second,
+		RandomizationFactor: 0.5,
+		MaxRetries:          5,
 	}
 	return http.Client{
 		Transport: MiddlewareRoundTripper{
@@ -44,7 +79,7 @@ func NewDefaultClient(metrics Metrics, roundTripper http.RoundTripper) http.Clie
 				metrics.ClientMiddleware,
 				jose.HTTPClientMiddleware,
 			},
-			RoundTripper: roundTripper,
+			RoundTripper: retryRoundTripper,
 		},
 	}
 }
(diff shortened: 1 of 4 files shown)


http/client.go
```diff
@@ -30,12 +33,44 @@ type MiddlewareRoundTripper struct {
...
+	RetriableStatusCodes []int
+	InitialInterval      time.Duration
+	RandomizationFactor  float64
+	Multiplier           float64
+	MaxInterval          time.Duration
+	MaxRetries           uint64
+}
+
 // NewDefaultClient constructs the default HTTP Client with middleware. Providing an HTTP
 // RoundTripper is optional. If `nil` is received, the DefaultClient will be used.
```
briancharous: Document the default configuration for retries

http/client.go
```diff
@@ -30,12 +33,44 @@ type MiddlewareRoundTripper struct {
...
+			http.StatusHTTPVersionNotSupported,
+			http.StatusVariantAlsoNegotiates,
+			http.StatusInsufficientStorage,
+			http.StatusLoopDetected,
+			http.StatusNotExtended,
+			http.StatusNetworkAuthenticationRequired,
+		},
+		InitialInterval:     100 * time.Millisecond,
+		Multiplier:          2,
+		MaxInterval:         30 * time.Second,
```
briancharous: This seems quite high, no?

http/client.go
```diff
@@ -30,12 +33,44 @@ type MiddlewareRoundTripper struct {
...
 
+// RetryRoundTripper wraps a roundtripper with retry logic
+type RetryRoundTripper struct {
+	RoundTripper         http.RoundTripper
+	RetriableStatusCodes []int
+	InitialInterval      time.Duration
+	RandomizationFactor  float64
+	Multiplier           float64
+	MaxInterval          time.Duration
+	MaxRetries           uint64
```
briancharous: probably nitpicky but seems like `uint8` would be fine here. I can't imagine wanting to retry more than 2^8 times.

http/client.go
```diff
@@ -30,12 +33,44 @@ type MiddlewareRoundTripper struct {
 	Middleware []ClientMiddleware
 }
 
+// RetryRoundTripper wraps a roundtripper with retry logic
+type RetryRoundTripper struct {
+	RoundTripper         http.RoundTripper
+	RetriableStatusCodes []int
```
briancharous: I guess this list is small, but should it be changed to a set?

http/client.go
```diff
@@ -30,12 +33,44 @@ type MiddlewareRoundTripper struct {
...
+
 // NewDefaultClient constructs the default HTTP Client with middleware. Providing an HTTP
 // RoundTripper is optional. If `nil` is received, the DefaultClient will be used.
 func NewDefaultClient(metrics Metrics, roundTripper http.RoundTripper) http.Client {
 	if roundTripper == nil {
 		roundTripper = http.DefaultTransport
 	}
+	retryRoundTripper := RetryRoundTripper{
+		RoundTripper: roundTripper,
+		RetriableStatusCodes: []int{
```
briancharous: It seems like 505-511 (`StatusHTTPVersionNotSupported` - `StatusNetworkAuthenticationRequired`) aren't actually retry-able.

http/client.go
benjigoldberg: I know, this is driven by the library wanting a uint64. I also laughed at this. I'll change it to uint8 and cast it internally.

http/client.go
benjigoldberg: Oh man, how did I not think of that 🤦‍♂  -- great call out, I'll update

http/client.go
benjigoldberg: Yep, good call.

http/client.go
benjigoldberg: well, its the minimum threshold of either up to a 30 second backoff or 5 retries, whichever is reached first. I could be talked into a lower interval. The default in this library is 15 minutes!

benjigoldberg: Updated to 10 seconds

briancharous: Ah ok. Maybe 30s is fine then. I was unsure if 30s was the total amount of time or the amount of time for one retry.

benjigoldberg: I'm wrong, this is the minimum threshold for any individual backoff, NOT the total. Leaving at 10s.

http/client.go
```diff
@@ -30,12 +33,45 @@ type MiddlewareRoundTripper struct {
...
+	MaxInterval          time.Duration
+	MaxRetries           uint8
+}
+
 // NewDefaultClient constructs the default HTTP Client with middleware. Providing an HTTP
 // RoundTripper is optional. If `nil` is received, the DefaultClient will be used.
+//
+// By default, the client is provides exponential backoff on 500-504 errors. The default
+// configuration for exponential backoff is to start with an interval of 100 milliseconds, a
+// multiplier of two, a randomization factor of 0.5 (for jitter), a max interval of 10 seconds,
```
wilsoniya: it's not clear to me from this description exactly how `RandomizationFactor` works.

http/client.go
```diff
@@ -86,3 +122,55 @@ func (mrt MiddlewareRoundTripper) RoundTrip(req *http.Request) (*http.Response,
...
+
+	var resp *http.Response
+	var err error
+	makeRequestRetriable := func() error {
+		resp, err = rrt.RoundTripper.RoundTrip(req)
+
+		// If an error was encountered, retry. This typically indicates a failure to get a
+		// response.
+		if err != nil {
+			log.Get(req.Context()).Debug("retrying failed http request", zap.Error(err))
```
wilsoniya: This log message seems misleading because this implementation will only retry if the status code is in `rrt.RetriableStatusCodes`.

benjigoldberg: @wilsoniya this actually is a retry on a failed request error. We return an err so the retry occurs here as well

PR #102, opened 2020-01-09
Feature/http cb

benjigoldberg:
Adds Hystrix Circuit-Breaking to our HTTP client.


http/client.go
```diff
@@ -123,3 +126,71 @@ func (rrt RetryRoundTripper) RoundTrip(req *http.Request) (*http.Response, error
 	}
 	return resp, err
 }
+
+// CircuitBreakerRoundTripper wraps a RoundTrapper with circuit-breaker logic
+type CircuitBreakerRoundTripper struct {
```
theisenmark: Do we need to insert an instance of this into the round tripper chain in `NewDefaultClient`? It would also be nice to add some documentation to explain the default config values, similar to the paragraph explaining exponential backoff.

http/client.go
```diff
@@ -123,3 +126,71 @@ func (rrt RetryRoundTripper) RoundTrip(req *http.Request) (*http.Response, error
...
+			RequestVolumeThreshold: hystrix.DefaultVolumeThreshold,
+			SleepWindow:            hystrix.DefaultSleepWindow,
+			ErrorPercentThreshold:  hystrix.DefaultErrorPercentThreshold,
+		},
+	}
+}
+
+// RoundTrip completes the http request round trip but wraps the call in circuit-breaking logic
+// using the Netflix Hystrix approach.
+func (cbrt CircuitBreakerRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
```
theisenmark: If possible, I'd prefer the circuit breaker functionality to be in its own file (`http/circuit_breaker.go`) similar to how the other middlewares are setup. I missed it before but the retry logic could be in its own file too.

http/client.go
```diff
@@ -123,3 +126,71 @@ func (rrt RetryRoundTripper) RoundTrip(req *http.Request) (*http.Response, error
...
+	// Ensure the RoundTripper was set on the CircuitBreakerRoundTripper
+	if cbrt.RoundTripper == nil {
+		panic("no roundtripper provided to circuit-breaker round tripper")
+	}
+
+	var resp *http.Response
+	var err error
+	makeRequestFunc := func(ctx context.Context) error {
+		resp, err = cbrt.RoundTripper.RoundTrip(req)
+		if err != nil || resp.StatusCode < http.StatusOK || resp.StatusCode >= http.StatusBadRequest {
```
briancharous: Shouldn't this be `>= http.StatusInternalServerError`? I don't think it's appropriate to circuit break on 4xx-level codes.

http/client.go
```diff
@@ -123,3 +126,71 @@ func (rrt RetryRoundTripper) RoundTrip(req *http.Request) (*http.Response, error
...
+			return fmt.Errorf("failed request, invoking circuit-breaker")
+		}
+		return nil
+	}
+
+	// Register the host configuration if it is not yet registered
+	if _, ok := cbrt.registeredHosts[req.URL.Host]; !ok {
```
briancharous: isn't this map going to need a mutex?

http/client.go
```diff
@@ -123,3 +126,71 @@ func (rrt RetryRoundTripper) RoundTrip(req *http.Request) (*http.Response, error
...
+// RoundTrip completes the http request round trip but wraps the call in circuit-breaking logic
+// using the Netflix Hystrix approach.
+func (cbrt CircuitBreakerRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	// Ensure the RoundTripper was set on the CircuitBreakerRoundTripper
+	if cbrt.RoundTripper == nil {
+		panic("no roundtripper provided to circuit-breaker round tripper")
+	}
+
+	var resp *http.Response
+	var err error
```
wilsoniya: I'm nervous that some person in the future making edits to this function might come along and reuse this `err` and thus disrupt how its intended to be used. Looks like `makeRequestFunc` has the side effect of setting `err`, so if someone comes along and sets `err` before `makeRequestFunc` is called, this function might end up erroneously returning a non-nil error. 

I'd recommend renaming this to `requestErr` or something.

http/metrics.go
```diff
@@ -178,26 +190,31 @@ func (metricsRT MetricsRoundTripper) RoundTrip(r *http.Request) (*http.Response,
...
 		}
-		metricsRT.metrics.clientCounter.With(labels).Inc()
-		if contentLengthStr := r.Header.Get("Content-Length"); len(contentLengthStr) > 0 {
-			if contentLength, err := strconv.Atoi(contentLengthStr); err == nil {
-				metricsRT.metrics.clientContentLength.With(labels).Observe(float64(contentLength))
+		if err != nil {
+			switch typedErr := err.(type) {
+			case hystrix.CircuitError:
+				if typedErr == hystrix.ErrCircuitOpen {
+					metricsRT.metrics.circuitBreakerOpen.With(prometheus.Labels{"host": r.URL.Host}).Inc()
```
wilsoniya: real nice

http/client.go
greg-mclain: To add to that, do we need the `< http.StatusOK` check? We shouldn't need to retry 1xx responses, right?

http/metrics.go
```diff
@@ -178,26 +190,31 @@ func (metricsRT MetricsRoundTripper) RoundTrip(r *http.Request) (*http.Response,
...
+			metricsRT.metrics.clientDuration.With(labels).Observe(durationSec)
 		}
-		metricsRT.metrics.clientCounter.With(labels).Inc()
-		if contentLengthStr := r.Header.Get("Content-Length"); len(contentLengthStr) > 0 {
-			if contentLength, err := strconv.Atoi(contentLengthStr); err == nil {
-				metricsRT.metrics.clientContentLength.With(labels).Observe(float64(contentLength))
+		if err != nil {
+			switch typedErr := err.(type) {
+			case hystrix.CircuitError:
+				if typedErr == hystrix.ErrCircuitOpen {
```
greg-mclain: Does this error also get returned when the circuit is already open? If so, this will count all of the failed requests while the circuit breaker is open.

http/client.go
benjigoldberg: 👍 great suggestions guys

http/metrics.go
benjigoldberg: Yes, my understanding is this will be returned until the circuit is closed again

benjigoldberg: @greg-mclain this is an excellent callout. I think we would be able to massage this into an on/off (open/closed) metric for alerting. I wonder if we're ok with that assumption here.

@wilsoniya @briancharous @theisenmark what do you guys think?

http/client.go
benjigoldberg: Good suggestion 👍

http/client.go
benjigoldberg: hah, yes

benjigoldberg:
@briancharous @wilsoniya @theisenmark @greg-mclain ready for another look. Thanks for the really great reviews guys! think you've all really helped prevent some heartache/bugs with your suggestions so far 👍

http/circuit_breaker.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+		return nil
+	}
+
+	// Register the host configuration if it is not yet registered
+	cbrt.mutex.RLock()
+	_, ok := cbrt.registeredHostsSet[req.URL.Host]
+	cbrt.mutex.RUnlock()
+	if !ok {
+		hystrixConfig, configExists := cbrt.HostConfiguration[req.URL.Host]
```
theisenmark: For my own education, why does `cbrt.registeredHostsSet[req.URL.Host]` above and below this get wrapped in a mutex but this `cbrt.HostConfiguration[req.URL.Host]` statement doesn't?

benjigoldberg: The reason I chose not to wrap the configuration with a mutex is an assumption that the configuration is a one-time setup item before we start to use the client. If we think callers were modifying the config map _after_ this time, it would be worth it to add a mutex.

Now that I'm writing this up, I think it may be worth it to mutex this as well, just in case.

http/metrics.go
greg-mclain: @benjigoldberg yea, I think we should be able to alert if the rate of `http_client_circuit_breaker_open_total` messages over the last X amount of time is >0. Could you tweak the Help text of this metric? It makes it sound like the counter is only incremented then the breaker opens.

http/circuit_breaker.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// NewDefaultCircuitBreakerRoundTripper constructs and returns the default
+// CircuitBreakerRoundTripper configuration.
+//
+// By default, circuit-breaking is configured on a host-by-host basis, meaning that errors for a
+// given host will be accumulated on that host for determining whether or not to open the circuit
+// breaker.
+//
+// The host configuration is a map of hostname to Hystrix configuration settings. Use of this
```
theisenmark: nit: `Hystrix` -> `circuit`

go.sum
```diff
@@ -16,6 +16,8 @@ github.com/Shopify/sarama v1.22.0 h1:rtiODsvY4jW6nUV6n3K+0gx/8WlAwVt+Ixt6RIvpYyo
 github.com/Shopify/sarama v1.22.0/go.mod h1:lm3THZ8reqBDBQKQyb5HB3sY1lKp3grEbQ81aWSgPp4=
 github.com/Shopify/toxiproxy v2.1.4+incompatible h1:TKdv8HiTLgE5wdJuEML90aBgNWsokNbMijUGhmcoBJc=
 github.com/Shopify/toxiproxy v2.1.4+incompatible/go.mod h1:OXgGpZ6Cli1/URJOF1DMxUHB2q5Ap20/P/eIdh4G0pI=
+github.com/afex/hystrix-go v0.0.0-20180502004556-fa1af6a1f4f5 h1:rFw4nCn9iMW+Vajsk51NtYIcwSTkXr+JGrMd36kTDJw=
```
briancharous: need to `go mod tidy` this away

http/circuit_breaker.go
```diff
@@ -0,0 +1,104 @@
...
+// **IMPORTANT**: If you decide to set the hostConfiguration, the key in the map **must be the host
+//                name of the server you intend to call (eg req.URL.Host)**
+//
+// Additional Default:
+// * Requests will timeout after 30 seconds
+// * No ceiling is placed on the number of concurrent requests per host (set to MaxInt32)
+// * 50% of requests must fail for the circuit-breaker to trip
+// * If the circuit-breaker opens, no requests will be attempted until after 5 seconds have passed
+// * At least 20 requests must be recorded to a host before the circuit breaker can be tripped
+func NewDefaultCircuitBreakerRoundTripper(
```
briancharous: awesome docs

http/circuit_breaker.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	// Fetch or register the circuit breaker for this host
+	circuitBreaker := cbrt.manager.GetCircuit(req.URL.Host)
+	if circuitBreaker == nil {
+		circuitBreaker = cbrt.manager.MustCreateCircuit(req.URL.Host)
+	}
+
+	var err error
+	if cbErr := circuitBreaker.Execute(req.Context(), makeRequestFunc, nil); cbErr != nil {
```
briancharous: Since I'm pretty sure we already require Go 1.13, you could eliminate all this logic and just wrap the error no matter what. Then in the various middlewares, unwrap it to avoid having to do a bunch of type assertions and not have to worry if something adds a new error along the way up the chain.

go.sum
benjigoldberg: idk why but its not going away

http/circuit_breaker.go
benjigoldberg: I want to do this as well, but with the way the actual types [are hidden in the circuit package](https://github.com/cep21/circuit/blob/master/v3/errors.go#L17) this is not currently possible.

go.sum
benjigoldberg: [wow, this package requires all of its competitors, apparently for benchmarks...](https://github.com/cep21/circuit/blob/fb2df49b5e0752ef2c00e47a4eb3eaa77d1498c2/v3/go.mod#L4)

benjigoldberg: See: https://github.com/cep21/circuit/issues/89

http/circuit_breaker.go
cep21: > with the way the actual types are hidden in the circuit package 

Can you clarify this.  Is there something you would want to improve?

benjigoldberg: Hi @cep21 -- thanks for asking! 

So in the errors file, [`errThrottledConcucrrentCommands`](https://github.com/cep21/circuit/blob/master/v3/errors.go#L5) and [`errCircuitOpen`] are unexported for us to use, as is [`circuitError`](https://github.com/cep21/circuit/blob/master/v3/errors.go#L9)

What we'd really like to be able to do is perform an [`errors.As(err, ErrCircuitOpen)`](https://godoc.org/errors#example-As). With the current implementation I can't access the underlying `errCircuitOpen` because its private to the circuit module.

I'm basing my understanding of this off this [Go blog post](https://blog.golang.org/go1.13-errors).

I hope that was helpful! Let me know if I've missed something or you'd like me to clarify.

cep21: It's purposeful that errors types are not exposed.  Dave Cheney has some great posts/thoughts on this here: https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully . The part to pay attention to is "Assert errors for behaviour, not type" and "Conclusion: avoid error types"

Does `errors.As` work if the second type is an interface?  Something like.

```
    var circuitError circuit.Error
    if errors.As(err, &circuit.Error) {
```

benjigoldberg: @cep21 thanks for the article, very informative. For what its worth your suggestion does absolutely work.

benjigoldberg: @cep21 thanks again for the information on this, you've taught us a lot today! All of us appreciate your feedback and information!

PR #108, opened 2020-02-06
add currency module with dollars to pennies conversion

benjigoldberg:
Adds a currency module with a simple dollars to pennies conversion


currency/conversions.go
```diff
@@ -0,0 +1,24 @@
+package currency
+
+import "math"
+
+// penniesToDollars represents that there are 100 pennies in a dollar
```
theisenmark: nit: `penniesToDollars` -> `penniesInDollar` to agree with the constant name

PR #115, opened 2020-03-18
rm auth required

benjigoldberg:
This gets rid of the concept of a globally enforced auth. We'll only ever enforce it on GRPC endpoints, HTTP routers, or HTTP Handlers. This was a broken idea that didn't work correctly.

### DIFF
--- a/jose/middleware.go
+++ b/jose/middleware.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 
 	"github.com/spothero/tools/log"
+	"go.uber.org/zap"
 )
 
 // authHeader defines the name of the header containing the JWT authorization data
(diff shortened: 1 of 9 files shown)


jose/middleware.go
```diff
@@ -31,64 +32,48 @@ const authHeader = "Authorization"
...
 			if len(authHeader) == 0 {
-				logger.Debug(authHeaderNotFound)
-				parseErrMsg = authHeaderNotFound
+				logger.Debug("no bearer token found")
+				next.ServeHTTP(w, r)
+				return
 			}
 
-			if len(parseErrMsg) == 0 && !strings.HasPrefix(authHeader, bearerPrefix) {
+			if !strings.HasPrefix(authHeader, bearerPrefix) {
```
wilsoniya: consider doing debug logging explicitly stating the request has not authorization header in the opposite side of this conditional.

PR #130, opened 2020-07-14
Add pagination package documentation

dj80hd:
I could not fine anywhere that [IDExtractor](https://godoc.org/github.com/spothero/tools/pagination#IDExtractor) is used.

If it is not used, should we remove it from the package?

### DIFF
--- a/pagination/example_test.go
+++ b/pagination/example_test.go
@@ -10,6 +10,7 @@ func (o Thing) ExtractID() string {
 	return string(o)
 }
 
+// ExamplePagination is an exmple
 func ExamplePagination() {
 	pages := []Pageable{
 		Thing("foo"),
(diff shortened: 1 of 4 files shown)


pagination/example_test.go
```diff
@@ -0,0 +1,26 @@
...
+  Thing("baz"),
+  Thing("boo"),
+  Thing("bla"),
+  }
+
+  page := GetPageAfterID(pages, "bar", 2)
+
+  fmt.Printf("%v", page)
+  //Output: [baz, boo]
+}
```
theisenmark, line 26: A pass through with `go fmt` would clean this up a bit.

dj80hd: good catch.  ran that on `paginator_test.go` too!

PR #141, opened 2020-09-10
feature/SEAR-920-increase-grpc-max-receive-msg-size

nbkincaid:
**Issue Link**
https://spothero.atlassian.net/browse/SEAR-920

**Description**
Valid search search transient availability responses were breaking the grpc message size default limit. This bumps it up to 20mb from roughly 4.2mb.

This also adds a format make target and commits the initial formatted codebase.


greg-mclain (approved):

.gitignore
```diff
@@ -15,3 +15,4 @@ example_server
 coverage.txt
 jwt-cli
 example-server
+.idea/*
```
dj80hd, line 18: is it common to check this in vs put it in one's global .gitignore?

Makefile
```diff
@@ -27,3 +27,6 @@ help: ## Prints this help command
 		sort | \
 		awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
+.PHONY: format
+format: ## formats the codebase
```
dj80hd, line 31: nice.  should lint depend on this?

.gitignore
nbkincaid: I'm not sure if we have a standard, but it seems to me its somewhat arbitrary in this case. Is there a downside you see?

Makefile
nbkincaid: I'm not sure, it depends on how we want to use it I'd say.

nbkincaid: I'm actually going to break this out into its own pr, I'm thinking the codecov run on master is behind and it's affecting the mergability of this pr.

http/circuit_breaker_test.go
```diff
@@ -48,7 +48,7 @@ func TestNewDefaultCircuitBreakerRoundTripper(t *testing.T) {
 			"host configuration is honored",
 			&mock.RoundTripper{ResponseStatusCodes: []int{http.StatusOK}, CreateErr: false},
 			map[string]circuit.Config{
-				"host": circuit.Config{},
+				"host": {},
```
dj80hd, line 51: oh cool, assignments to slice and map values have an implicit type!

PR #158, opened 2020-12-11
add support for global dynamic log level

greg-mclain:
**Description**
Adds the ability to change the global log level of an application.

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"sync"
 
+	"github.com/gorilla/mux"
 	grpc_zap "github.com/grpc-ecosystem/go-grpc-middleware/logging/zap"
 	"github.com/prometheus/client_golang/prometheus"
 	"go.uber.org/zap"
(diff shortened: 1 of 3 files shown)


log/log.go
```diff
@@ -147,3 +151,9 @@ func Get(ctx context.Context) *zap.Logger {
 	}
 	return logger
 }
+
+// RegisterLogLevelHandler registers an endpoint handler with the specified router to change the global log level
+// for the application.
+func RegisterLogLevelHandler(router *mux.Router) {
+	router.Handle("/loglevel", globalLogLevel)
```
theisenmark, line 158: This is super cool! Can we add a bit more documentation around this? Maybe here or in the README or something? I had to dive into the [zap docs](https://github.com/uber-go/zap/blob/master/http_handler.go#L34-L36) to figure out exactly how to use this endpoint.

theisenmark (approved):

dj80hd (approved):
Very cool

log/log.go
greg-mclain: Good suggestion. I added it in the function documentation for now, although this is a reminder that we don't have great documentation in general on how to use several of the tools here. We should circle back and either fill out the root level README better and/or add READMEs for each of the components.

PR #159, opened 2021-01-13
Implement schema registry enhancements

ryan-carlson:
**Issue Link**
https://spothero.atlassian.net/browse/SIQ-328

**Description**
This change enhances the schema registry client to add support for new APIs:
- CheckSchema: checks if a schema is exists in the registry
- CreateSchema: creates a new schema in the registry

### DIFF
--- a/kafka/schema_registry.go
+++ b/kafka/schema_registry.go
@@ -15,6 +15,7 @@
 package kafka
 
 import (
+	"bytes"
 	"context"
 	"encoding/binary"
 	"encoding/json"
@@ -37,6 +38,22 @@ type SchemaRegistryConfig struct {
 	URL string
 }
 
+type schemaRequest struct {
+	Schema string `json:"schema"`
+}
+
+type schemaResponse struct {
+	Subject string `json:"subject"`
+	Version int    `json:"version"`
+	Schema  string `json:"schema"`
+	ID      int    `json:"id"`
+}
+
+type errorResponse struct {
+	ErrorCode int `json:"error_code"`
+	Message string `json:"message"`
+}
+
 // RegisterFlags registers schema registry flags with pflags
 func (c *SchemaRegistryConfig) RegisterFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&c.URL, "kafka-schema-registry-url", "http://localhost:8081", "Kafka schema registry url")
@@ -50,7 +67,8 @@ func (c *SchemaRegistryConfig) RegisterFlags(flags *pflag.FlagSet) {
 type SchemaRegistryClient struct {
 	SchemaRegistryConfig
 	client http.Client
-	cache  *sync.Map
+	cache  sync.Map
+	cacheLock sync.RWMutex
 }
 
 // NewSchemaRegistryClient creates a schema registry client with the given HTTP metrics bundle.
@@ -75,7 +93,7 @@ func (c SchemaRegistryConfig) NewSchemaRegistryClient(httpMetrics shHTTP.Metrics
 	return SchemaRegistryClient{
 		SchemaRegistryConfig: c,
 		client:               http.Client{Transport: metricsRoundTripper},
-		cache:                &sync.Map{},
+		cache:                sync.Map{},
 	}
 }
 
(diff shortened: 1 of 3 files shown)


philipp-birklbauer (commented):
Have two questions. Looks good otherwise I think

kafka/schema_registry.go
```diff
@@ -50,7 +67,8 @@ func (c *SchemaRegistryConfig) RegisterFlags(flags *pflag.FlagSet) {
 type SchemaRegistryClient struct {
 	SchemaRegistryConfig
 	client http.Client
-	cache  *sync.Map
+	cache  sync.Map
```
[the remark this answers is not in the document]
greg-mclain: Another option is to keep it as a value and return a pointer to `SchemaRegistryClient` from `NewSchemaRegistryClient`. We've used both patterns throughout our other Go services and I'm not sure if one approach is better than the other.

kafka/schema_registry.go
```diff
@@ -98,52 +117,166 @@ func (c SchemaRegistryClient) GetSchema(ctx context.Context, id uint) (string, e
...
 }
 
-// DecodeKafkaAvroMessage decodes the given Kafka message encoded with Avro into a Go type.
-func (c SchemaRegistryClient) DecodeKafkaAvroMessage(ctx context.Context, message *sarama.ConsumerMessage) (interface{}, error) {
-	// bytes 1-4 are the schema id (big endian), bytes 5... is the message
-	// see: https://docs.confluent.io/current/schema-registry/docs/serializer-formatter.html#wire-format
-	if len(message.Value) < 5 {
-		return nil, fmt.Errorf("no schema id not found in Kafka message")
+// CreateSchema creates a new schema in Schema Registry
+func (c *SchemaRegistryClient) CreateSchema(ctx context.Context, subject string, schema string, isKey bool) (*schemaResponse, error) {
```
greg-mclain, line 179: Out of curiosity, what happens if a client calls `CreateSchema` for a schema that already exists?

kafka/schema_registry.go
```diff
@@ -98,52 +117,166 @@ func (c SchemaRegistryClient) GetSchema(ctx context.Context, id uint) (string, e
...
+		}
+		return nil, fmt.Errorf("%s, error code %d", errorResponse.Message, errorResponse.ErrorCode)
+	default:
+		return nil, fmt.Errorf(
+			"error while creating schema; schema registry returned unhandled status code %d", resp.StatusCode)
+	}
+}
+
+// GetCodec returns an avro codec based on the provided schema id
+func (c *SchemaRegistryClient) GetCodec(ctx context.Context, id uint) (*goavro.Codec, error) {
```
greg-mclain, line 225: Does this method need to be exported?

kafka/schema_registry.go
```diff
@@ -50,7 +67,8 @@ func (c *SchemaRegistryConfig) RegisterFlags(flags *pflag.FlagSet) {
 type SchemaRegistryClient struct {
 	SchemaRegistryConfig
 	client http.Client
-	cache  *sync.Map
+	cache  sync.Map
+	cacheLock sync.RWMutex
```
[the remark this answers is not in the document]
ryan-carlson: Damn, I overlooked that.  I have pushed a change to remove the Mutex.

kafka/schema_registry.go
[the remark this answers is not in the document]
ryan-carlson: I like the second option better so I have made the update.

I am not expecting users to directly access the cache (this is meant to be private).  All access should be done through the `GetCodec` function.  And I am assuming since it is a pointer receiver on `GetCodec` that no copies are created when accessing the `c.cache`.

kafka/schema_registry.go
ryan-carlson: The schema registry will compare the schema.  If it finds this in the registry, it will return the existing ID.  

Note that I realized this only after creating the CheckSchema function.  I don't think we really need this since CreateSchema checks it anyways.

kafka/schema_registry.go
ryan-carlson: Good question, I wasn't sure if users would need to interact directly with the Codec.  I am expecting most usage to be using the `DecodeKafkaAvroMessage`.  But I left it since I didn't see a reason not to expose this functionality.

kafka/schema_registry.go
ryan-carlson: Note:  I went ahead and update the comment to provide this information.

greg-mclain (approved):

philipp-birklbauer (approved):

PR #160, opened 2021-01-15
Enhance schema registry support for encoding avro messages

ryan-carlson:
**Issue Link**
N/A

**Description**
This change adds a new API to the schema registry client that will encode avro messages from a golang map.


philipp-birklbauer (approved):
LGTM

PR #162, opened 2021-01-15
Handle AlreadyRegisteredError when creating new http metrics

ryan-carlson:
**Issue Link**
N/A

**Description**
This change will handle a case where AlreadyRegisteredError is returned from Prometheus when registering a collector.  In this case, the existing collector will be returned.

### DIFF
--- a/http/metrics.go
+++ b/http/metrics.go
@@ -35,13 +35,37 @@ type Metrics struct {
 	circuitBreakerOpen  *prometheus.CounterVec
 }
 
+// registerCollector will register the passed collector
+func registerCollector(registry prometheus.Registerer, collector prometheus.Collector, mustRegister bool) prometheus.Collector {
+	if mustRegister {
+		registry.MustRegister(collector)
+	} else {
+		if err := registry.Register(collector); err != nil {
+			if are, ok := err.(prometheus.AlreadyRegisteredError); ok {
+				// metric has been registered before so use existing metric
+				return are.ExistingCollector
+			} else {
+				// Something else went wrong so panic
+				panic(err)
+			}
+		}
+	}
+	return collector
+}
+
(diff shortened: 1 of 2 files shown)


http/metrics.go
```diff
@@ -35,13 +35,37 @@ type Metrics struct {
...
+	if mustRegister {
+		registry.MustRegister(collector)
+	} else {
+		if err := registry.Register(collector); err != nil {
+			if are, ok := err.(prometheus.AlreadyRegisteredError); ok {
+				// metric has been registered before so use existing metric
+				return are.ExistingCollector
+			} else {
+				// Something else went wrong so panic
+				panic(err)
```
ryan-carlson, line 49: @greg-mclain 
This is a potentially breaking change.  When this happened previously, the logic would silently ignore this even though nothing had been registered.

greg-mclain: I'm ok with that. I don't think we'll have issues and if we do they'll be easy to track down.

http/metrics.go
```diff
@@ -35,13 +35,37 @@ type Metrics struct {
...
+		}
+	}
+	return collector
+}
+
 // NewMetrics creates and returns a metrics bundle. The user may optionally
 // specify an existing Prometheus Registry. If no Registry is provided, the global Prometheus
 // Registry is used. Finally, if mustRegister is true, and a registration error is encountered,
-// the application will panic.
+// the application will panic.  If mustRegister is false,
```
greg-mclain, line 59: The last sentence is unfinished.

greg-mclain (approved):
Thanks for doing this!

http/metrics.go
ryan-carlson: Fixed, I added a more in depth comment to explain exactly what to expect.

PR #169, opened 2021-06-01
time-delta-helpers

noahwill:
**Description**
In this PR, I add some helpers in the `time` package to make incrementing/decrementing different units of time easier.

### DIFF
--- a/time/time.go
+++ b/time/time.go
@@ -15,6 +15,7 @@
 package time
 
 import (
+	"fmt"
 	"sync"
 	"time"
 )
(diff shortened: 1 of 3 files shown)


time/time.go
```diff
@@ -47,3 +48,41 @@ func LoadLocation(name string) (*time.Location, error) {
...
+}
+
+// IncDay increments time t by count number of days
+func IncDay(t time.Time, count int) time.Time {
+	return t.AddDate(0, 0, 1*count)
+}
+
+// IncHour increments time t by count number of hours
+func IncHour(t time.Time, count int) time.Time {
+	d, _ := time.ParseDuration(fmt.Sprintf("%vh", count))
```
theisenmark, line 74: Here's another way to accomplish the same thing without needing to swallow an error:
```
func IncHour(t time.Time, count int) time.Time {
  return t.Add(time.Hour * count)
}
```

Although with an implementation that simple I'm not sure we need a helper at all.

theisenmark (commented):
Let's add some tests that cross DST boundaries to make sure we're covered.

time/time.go
greg-mclain: I agree with Mark. It may not be worth it if it's a one-liner like that. Also, there is a helper function in periodic that adds DST aware durations to a time: https://github.com/spothero/periodic/blob/main/periodic.go#L300-L308 I'm not sure if it makes most sense to use that for your use cases @noahwill, potentially wrap it with some more declarative functions as you have here, or if DST doesn't need to be considered.

noahwill: Ah good catch - the week, minute, and second cases are the only ones that are annoying to do in line, the others can be one line but are included mostly for consistency

noahwill: actually, now that I went to change it I see why the hours one was annoying as well - `time.Hour` is a duration and `count` is an int so you get a type error when trying to multiply. If it were `time.Hour * 1` (just that constant value) then there would be no error. Like I said, some of the ones that could be omitted since they are just one line are included for consistency in the library, so I think leaving this in is fine.

theisenmark: Oof yeah that's annoying, I guess you'd have to wrap it then. Still not so bad.
```
func IncHour(t time.Time, count int) time.Time {
  return t.Add(time.Hour * time.Duration(count))
}
```

noahwill:
@greg-mclain pointed me to some appropriate functionality in the periodic lib

PR #170, opened 2021-07-15
SEAR-1541 Add option to disable Sentry integration

theisenmark:
**Issue Link**
https://spothero.atlassian.net/browse/SEAR-1541

**Description**
Adds option to disable Sentry integration. Sentry is still enabled by default.


noahwill (approved):
✅

PR #172, opened 2021-08-13
Format existing source code

ryan-carlson:
**Issue Link**
NA

**Description**
I found that some of the existing source files were not formatted when development the authorization by scope feature.  This PR will format these files to bring them up to the standard.


theisenmark (approved):
Thanks for cleaning this up!

PR #183, opened 2022-04-27
allow client to specify default config for service [DO NOT MERGE]

dj80hd:


dj80hd:
This is just too messy to use.

PR #186, opened 2022-06-09
Upgrade to Go 1.18

theisenmark:
**Description**
Upgraded to `Go 1.18` and ran go mod tidy to clean up dependencies.


grpc/client.go
```diff
@@ -46,7 +47,7 @@ type ClientConfig struct {
 // NewDefaultClientConfig returns the default SpotHero gRPC plaintext Client Configuration
 func NewDefaultClientConfig(ctx context.Context) ClientConfig {
 	cc := defaultClientConfig(ctx)
-	cc.Options = append(cc.Options, grpc.WithInsecure())
```
theisenmark, line 49: `grpc.WithInsecure()` is marked as deprecated - this should be an equivalent replacement.

kafka/producer_test.go
```diff
@@ -55,23 +55,23 @@ func TestAsyncProducer_run(t *testing.T) {
...
+				Value: sarama.ByteEncoder([]byte("value")),
+				Topic: "topic",
 			}
 			producer.Input() <- msg
 
-			labels := prometheus.Labels{"topic": "topic", "partition": "0"}
 			if test.fail {
 				msgErr := <-producer.Errors()
 				assert.Equal(t, msg, msgErr.Msg)
+				labels := prometheus.Labels{"topic": "topic", "partition": fmt.Sprint(msg.Partition)}
```
theisenmark, line 66: Sarama's mock producer no longer guarantees which partition gets used for messages. We're dynamically filling it in here since it's not a critical piece of the test.

noahwill (approved):

PR #198, opened 2022-10-18
SEAR-2281: Migrating tracing from OpenTracing to OpenTelemetry.

abhioncbr:
**Issue Link**
[SEAR-2281](https://spothero.atlassian.net/browse/SEAR-2281)

**Description**
- Migrating tracing from OpenTracing to OpenTelemetry.


sentry/middleware.go
```diff
@@ -26,7 +25,7 @@ func NewMiddleware() Middleware {
 	}
 }
 
-// This middleware is a wrapper around the sentry-go library's middleware that attaches a the Sentry hub that
+// HTTP This middleware is a wrapper around the sentry-go library's middleware that attaches a the Sentry hub that
```
theisenmark, line 28: nit: I'm assuming this was added because Go always wants you to begin the comment with the function name? If so, I think it reads better as just "HTTP is a wrapper around..."

service/service.go
```diff
@@ -125,19 +125,28 @@ func (c Config) ServerCmd(
...
+
+			shutdown, err := tc.TracerProvider()
+			if err != nil {
+				return err
+			}
+			defer func() {
+				err = shutdown(ctx)
+			}()
+			if err != nil {
+				return err
```
theisenmark, line 137: Should this be in the `defer` block above? Otherwise it doesn't seem like it's doing much here

tracing/middleware.go
```diff
@@ -143,16 +140,17 @@ func SQLMiddleware(ctx context.Context, queryName, query string, args ...interfa
...
-		SetTag("component", "tracing").
-		SetTag("db.type", "sql").
-		SetTag("db.statement", query).
-		SetTag("db.statement.arguments", args)
+	span, spanCtx := StartSpanFromContext(ctx, spanName)
+	var attrs []attribute.KeyValue
+	attrs = append(attrs, attribute.String("component", "tracing"))
+	attrs = append(attrs, attribute.String("db.type", "sql"))
+	attrs = append(attrs, attribute.String("db.statement", query))
+	//attrs = append(attrs, attribute.StringSlice("db.statement.arguments", args))
```
theisenmark, line 148: Should this still be disabled?

tracing/tracing.go
```diff
@@ -51,61 +52,71 @@ type Config struct {
...
-}
 
-// TraceOutbound injects outbound HTTP requests with OpenTracing headers
-func TraceOutbound(r *http.Request, span opentracing.Span) error {
-	return opentracing.GlobalTracer().Inject(
-		span.Context(),
-		opentracing.HTTPHeaders,
-		opentracing.HTTPHeadersCarrier(r.Header))
+	// Set sampler for the traceprovider
+	sampler := tracesdk.AlwaysSample()
```
theisenmark, line 81: So the sampler properties will no longer be configurable?

tracing/middleware.go
abhioncbr: The attribute in OTEL doesn't support interface slices. It requires an exact data type as an argument.

abhioncbr:
@theisenmark Can you please review it again? Thanks.

noahwill (approved):
LGTM!

tracing/middleware.go
theisenmark: Gotcha. I'd say let's just delete it then.

tracing/tracing.go
theisenmark: @abhioncbr ^

abhioncbr: It's configurable. Same as before, however, based on types.

The sampler is based on type. currently supporting three types
- `Always` (default), not suggested for production workload
- `Ratio-based`- Need another param for configuring the ratio. 1 or greater then 1 means Always & 0 or less then 0 means Never.
- `Neve`r.
We need to update the tracer sampler property in app charts.

theisenmark (approved):

PR #200, opened 2023-02-09
SEAR-2441 track req received before return; add req returned metric

noahwill:
**Issue Link**
https://spothero.atlassian.net/browse/SEAR-2441

**Description**
This PR moves our request counter metric to increment when the request is received and adds a new requests returned metric.

### DIFF
--- a/http/metrics.go
+++ b/http/metrics.go
@@ -31,6 +31,7 @@ const UNAUTHENTICATED = "unauthenticated"
 // Metrics is a bundle of prometheus HTTP metrics recorders
 type Metrics struct {
 	counter             *prometheus.CounterVec
+	requestReturned     *prometheus.CounterVec
 	duration            *prometheus.HistogramVec
 	contentLength       *prometheus.HistogramVec
 	clientCounter       *prometheus.CounterVec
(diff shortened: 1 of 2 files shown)


http/metrics.go
```diff
@@ -97,14 +110,14 @@ func NewMetrics(registry prometheus.Registerer, mustRegister bool) Metrics {
 	)
 	clientHistogram = registerCollector(registry, clientHistogram, mustRegister).(*prometheus.HistogramVec)
 
-	counter := prometheus.NewCounterVec(
+	requestReturnedCounter := prometheus.NewCounterVec(
 		prometheus.CounterOpts{
-			Name: "http_requests_total",
-			Help: "Total number of HTTP Requests received",
+			Name: "http_requests_returned_total",
```
theisenmark, line 115: Does `http_responses_total` make more sense? Or maybe `http_responses_returned_total`? `http_requests_completed_total`? It doesn't actually matter, feel free to ignore this - just something about "requests returned" feels a bit contradictory

theisenmark (approved):

http/metrics.go
noahwill: I like `http_responses_total` 👍🏻

http/metrics.go
```diff
@@ -162,15 +176,17 @@ func NewMetrics(registry prometheus.Registerer, mustRegister bool) Metrics {
 // for HTTP response code tagging to function.
 func (m Metrics) Middleware(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		labels := prometheus.Labels{
+			"path":                 writer.FetchRoutePathTemplate(r),
+			"authenticated_client": retrieveAuthenticatedClient(r),
+		}
+		m.counter.With(labels).Inc()
```
nbkincaid, line 183: Can we call this requestCounter just for the sake of following convention with the responseCounter?

PR #232, opened 2024-06-05
[MOB-477] Fixes the Auth0Claims for Refresh token

amitc005:
**Issue Link**
MOB-477

**Description**
- This Pr fixes the error when `jwt` was trying to unmarshal `gty` auth0 custom claims. In Auth0, when the token is get from the /token API, it comes as a string, but the token from the refresh token comes as an array, i.e ["refresh_token", "password"]. Therefore this PR makes the grant type field an interface in Auth0Claim

### DIFF
--- a/jose/auth0.go
+++ b/jose/auth0.go
@@ -36,11 +36,11 @@ type Auth0Generator struct{}
 type Auth0Claim struct {
 	ID string `json:"sub"`
 	// Email claims are namespaced to prevent collisions. Hardcoding for now as this will be constant.
-	Email          string `json:"https://api.spothero.com/claims/email"`
-	ClientName     string `json:"https://api.spothero.com/claims/clientName"`
-	GrantType      string `json:"gty"`
-	Scope          string `json:"scope"`
-	ExternalUserID string `json:"https://api.spothero.com/claims/user_id"`
+	Email          string      `json:"https://api.spothero.com/claims/email"`
+	ClientName     string      `json:"https://api.spothero.com/claims/clientName"`
+	GrantType      interface{} `json:"gty"`
+	Scope          string      `json:"scope"`
+	ExternalUserID string      `json:"https://api.spothero.com/claims/user_id"`
 }
 
 // New satisfies the ClaimGenerator interface, returning an empty claim for use with JOSE parsing
--- a/jose/auth0.go
+++ b/jose/auth0.go
@@ -58,7 +58,7 @@ func (cc Auth0Claim) NewContext(ctx context.Context) context.Context {
 // GetClientID returns the ClientID field of the claim if it is present,
 // otherwise the empty string
 func (cc Auth0Claim) GetClientID() string {
-	if reflect.TypeOf(cc.GrantType).Kind() == reflect.Slice {
+	if reflect.TypeOf(cc.GrantType).Kind() == reflect.Slice && reflect.TypeOf(cc.GrantType).Elem().Kind() == reflect.String {
 		for _, grantType := range cc.GrantType.([]string) {
 			if grantType == "client-credentials" {
 				// because Auth0 adds the undesirable suffix of "@clients"
@@ -78,7 +78,7 @@ func (cc Auth0Claim) GetClientID() string {
 // GetUserID returns the UserID field of the claim if it is present, otherwise
 // the empty string
 func (cc Auth0Claim) GetUserID() string {
-	if reflect.TypeOf(cc.GrantType).Kind() == reflect.Slice {
+	if reflect.TypeOf(cc.GrantType).Kind() == reflect.Slice && reflect.TypeOf(cc.GrantType).Elem().Kind() == reflect.String {
 		for _, grantType := range cc.GrantType.([]string) {
 			if grantType == "password" || grantType == "authorization_code" {
 				return cc.ID
--- a/jose/auth0.go
+++ b/jose/auth0.go
@@ -17,6 +17,7 @@ package jose
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


amitc005:
Refresh token

```JSON
{
  "https://api.spothero.com/claims/email": "amitc005@gmail.com",
  "https://api.spothero.com/claims/user_id": "cd4fd237-ee87-4ecf-b1bc-71850c8ee7eb",
  "https://api.spothero.com/claims/email_verified": true,
  "https://api.spothero.com/claims/created_at": "2023-02-10T21:09:49.032Z",
  "https://api.spothero.com/claims/monolith_originated": true,
  "https://api.spothero.com/claims/clientName": "SpotHero-iOS",
  "iss": "https://accounts.stg.spothero.com/",
  "sub": "apple|000661.828f5b1a15e6403bbc462d81fc20911b.2115",
  "aud": [
    "https://api.staging.spothero.com/v2/",
    "https://spothero-stg.us.auth0.com/userinfo"
  ],
  "iat": 1717556473,
  "exp": 1717642873,
  "scope": "openid email offline_access",
  "gty": [
    "refresh_token",
    "password"
  ],
  "azp": "uxmnxFdVc8hJxdClQcB7YdiOlPukn744"
}
```

jose/auth0.go
```diff
@@ -36,11 +36,11 @@ type Auth0Generator struct{}
...
 	ID string `json:"sub"`
 	// Email claims are namespaced to prevent collisions. Hardcoding for now as this will be constant.
-	Email          string `json:"https://api.spothero.com/claims/email"`
-	ClientName     string `json:"https://api.spothero.com/claims/clientName"`
-	GrantType      string `json:"gty"`
-	Scope          string `json:"scope"`
-	ExternalUserID string `json:"https://api.spothero.com/claims/user_id"`
+	Email          string      `json:"https://api.spothero.com/claims/email"`
+	ClientName     string      `json:"https://api.spothero.com/claims/clientName"`
+	GrantType      interface{} `json:"gty"`
```
noahwill, line 41: Do you think it's worth having this be `[]string` or are we just passing it on and don't need to worry about typing it?

theisenmark (commented):
`GrantType` is used in a [couple places](https://github.com/spothero/tools/blob/9a840a8e5fb5c0292ee336fe5ae64cd277ce1b0c/jose/auth0.go#L57-L76) throughout the code - let's make sure those methods continue to function as expected after this change. In particular, we should still be able to use `GetUserID` to extract an ID from a token even when `gty` is an array.

amitc005:
@theisenmark @noahwill is there an alternative way to resolve this problem? Or the one I proposed in PR is acceptable?

jose/auth0.go
amitc005: @noahwill does it support just string as well? for token from `/token` sends `gty` as `string`

amitc005:
> `GrantType` is used in a [couple places](https://github.com/spothero/tools/blob/9a840a8e5fb5c0292ee336fe5ae64cd277ce1b0c/jose/auth0.go#L57-L76) throughout the code - let's make sure those methods continue to function as expected after this change. In particular, we should still be able to use `GetUserID` to extract an ID from a token even when `gty` is an array.

https://github.com/spothero/tools/pull/232/commits/e394dc40da076e652f046c92bc3f68aeff14c6dc

jose/auth0.go
```diff
@@ -57,6 +58,15 @@ func (cc Auth0Claim) NewContext(ctx context.Context) context.Context {
 // GetClientID returns the ClientID field of the claim if it is present,
 // otherwise the empty string
 func (cc Auth0Claim) GetClientID() string {
+	if reflect.TypeOf(cc.GrantType).Kind() == reflect.Slice {
+		for _, grantType := range cc.GrantType.([]string) {
```
abhioncbr, line 62: I think to make it less error-prone, check the type of element of the slice as well before converting to `[]string`.
you can use something like this 
```
reflect.TypeOf(cc.GrantType).Elem() == string
```

amitc005: @abhioncbr with the existing check so something like this?

```GO
reflect.TypeOf(cc.GrantType).Kind() == reflect.Slice && reflect.TypeOf(cc.GrantType).Elem().Kind() == reflect.String
```

amitc005: Done: [8dcf615](https://github.com/spothero/tools/pull/232/commits/8dcf615c07d76bed8dd3256515f4a3958c111ef2)

abhioncbr: It looks good to me on this check. Please make sure to address comments as well. I am approving the pr.

abhioncbr (approved):

jose/auth0.go
noahwill: ah yeah ok that makes sense to have it be an interface then if it can be two types

jose/auth0_test.go
```diff
@@ -70,18 +70,33 @@ func TestGetClientID(t *testing.T) {
 	tests := []struct {
 		name     string
 		input    Auth0Claim
-		expected string
+		expected interface{}
```
theisenmark, line 73: No need to change this, the return type will still always be `string`

jose/auth0_test.go
```diff
@@ -106,13 +121,14 @@ func TestGetUserID(t *testing.T) {
...
```
theisenmark, line 124: No need to change this, the return type will still always be `string`

theisenmark (approved):

delannoyk:
Hey @amitc005! Any estimate as to when this will get merged in and deployed?

jose/auth0_test.go
amitc005: @theisenmark Sorry I did not see your comment I will create another PR.

PR #234, opened 2024-06-11
ADHOC - Handle Auth0 claims with nil grant types

theisenmark:
**Issue Link**
ADHOC

**Description**
Handles Auth0 claims with `GrantType == nil`

### DIFF
--- a/jose/auth0.go
+++ b/jose/auth0.go
@@ -58,7 +58,7 @@ func (cc Auth0Claim) NewContext(ctx context.Context) context.Context {
 // GetClientID returns the ClientID field of the claim if it is present,
 // otherwise the empty string
 func (cc Auth0Claim) GetClientID() string {
-	if reflect.TypeOf(cc.GrantType).Kind() == reflect.Slice && reflect.TypeOf(cc.GrantType).Elem().Kind() == reflect.String {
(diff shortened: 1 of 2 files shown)


amitc005 (approved):

jose/auth0.go
```diff
@@ -58,7 +58,7 @@ func (cc Auth0Claim) NewContext(ctx context.Context) context.Context {
 // GetClientID returns the ClientID field of the claim if it is present,
 // otherwise the empty string
 func (cc Auth0Claim) GetClientID() string {
-	if reflect.TypeOf(cc.GrantType).Kind() == reflect.Slice && reflect.TypeOf(cc.GrantType).Elem().Kind() == reflect.String {
+	if cc.GrantType != nil && reflect.TypeOf(cc.GrantType).Kind() == reflect.Slice && reflect.TypeOf(cc.GrantType).Elem().Kind() == reflect.String {
```
abhioncbr, line 61: Since we have converted it to a string instead of an interface, we don't need to check elem kind as a string.

theisenmark: `GrantType` is still an [interface](https://github.com/spothero/tools/blob/91a63b6c3c66541c2a477d443649d9cfee76dad4/jose/auth0.go#L42). Shouldn't we continue checking that the elements are strings here?

PR #238, opened 2024-08-29
Bump golang to 1.22.6

twessels-sh:
**Issue Link**

Ad hoc

**Description**

I noticed that some of our other repos were on go v.1.22.6, so this PR updates its version of go to 1.22.6.


jose/middleware.go
```diff
@@ -161,13 +162,13 @@ func validateRequiredScope(r *http.Request, params AuthParams) error {
 	if params.RequiredScopes != nil {
 		claim, err := FromContext(r.Context())
 		if err != nil {
-			return fmt.Errorf(cannotFindClaim)
+			return errors.New(cannotFindClaim)
```
twessels-sh, line 165: Changes are like these are due to some linting errors:

```
golangci-lint run
service/cli.go:63:21: printf: non-constant format string in call to fmt.Errorf (govet)
                return fmt.Errorf(strings.Join(errors, ", "))
                                  ^
jose/middleware.go:164:11: SA1006: printf-style function with dynamic format string and no further arguments should use print-style function instead (staticcheck)
                        return fmt.Errorf(cannotFindClaim)
                               ^
jose/middleware.go:170:12: SA1006: printf-style function with dynamic format string and no further arguments should use print-style function instead (staticcheck)
                                return fmt.Errorf(missingRequiredScope)
                                       ^
jose/middleware_test.go:324:14: SA1006: printf-style function with dynamic format string and no further arguments should use print-style function instead (staticcheck)
                        expected: fmt.Errorf(cannotFindClaim),
                                  ^
make: *** [Makefile:20: lint] Error 1

Exited with code exit status 2
```

I'm new to Go, so let me know if I correctly addressed these.

theisenmark (approved):

PR #239, opened 2024-09-04
Export traces to OTel collector instead of Jaeger

twessels-sh:
**Description**

Adds an OpenTelemetry (OTel) [OTLP exporter](https://opentelemetry.io/docs/languages/go/exporters/#otlp) to send traces to the OTel collector that is running in each of our environment clusters. The Jaeger exporter is removed, as the OTel collector is configured to forward tracing data to Jaeger.

We already export traces directly to Jaeger, so why introduce this exporter? The primary driver is to get the data to Sumo Logic so we can trial their [APM offering](https://help.sumologic.com/docs/apm/). But more importantly, this is the pattern that we ultimately want to adopt (and that some other services have already adopted), where tracing data is sent to the OTel collector, with the collector being the single place to control how data is gathered, processed, and exported.

As part of this work, the `deployment.environment` attribute was added to traces , while the value for the `service.version` attribute was modified.

**Testing**

Used `go mod edit -replace` to have my local instance of the availability service use my local version of this module, and then validated that traces produced by the availability service were exported to a local OTel collector and included the updated attributes.


grpc/client.go
```diff
@@ -106,7 +106,7 @@ func (cc ClientConfig) GetConn() (*grpc.ClientConn, error) {
 		cc.UnaryInterceptors = append(cc.UnaryInterceptors, grpcretry.UnaryClientInterceptor(opts...))
 		cc.StreamInterceptors = append(cc.StreamInterceptors, grpcretry.StreamClientInterceptor(opts...))
 	}
-	return grpc.Dial(
+	return grpc.NewClient(
```
twessels-sh, line 109: Running `go mod tidy` resulted in the `grpc` library being updated to a new version, and this new version has `Dial` deprecated. I made the suggested change - use `NewClient` instead - to pass linting. 

According to [this](https://github.com/grpc/grpc-go/blob/master/Documentation/anti-patterns.md#the-wrong-way-grpcdial),`NewClient` has some functional differences compared to `Dial`. And given this, I'm wary of including this change in my PR, as I'm new to the Go ecosystem and our Go-based services, and the change at least _feels_ like it merits closer attention by someone more experienced. If I'm overestimating the impact of the change or if testing this change is straightforward, I'm fine with keeping this change in the PR. But otherwise, is there a way to revert the `grpc` library version? My attempts to use an older version of the library kept being overwritten with the new version.

tracing/tracing.go
```diff
@@ -96,19 +94,19 @@ func (c Config) TracerProvider() (func(context.Context) error, error) {
 		semconv.SchemaURL,
 		semconv.ServiceNameKey.String(c.ServiceName),
 		semconv.ServiceNamespaceKey.String(c.ServiceNamespace),
-		semconv.ServiceVersionKey.String(os.Getenv("VERSION")),
+		semconv.ServiceVersionKey.String(c.ServiceVersion),
+		semconv.DeploymentEnvironmentKey.String(c.Environment),
 		semconv.TelemetrySDKLanguageGo,
 		semconv.TelemetrySDKNameKey.String("opentelemetry"),
-		semconv.TelemetrySDKVersionKey.String("1.23.1"),
+		semconv.TelemetrySDKVersionKey.String("1.29.0"),
```
twessels-sh, line 101: Not crazy with having to manually set this value, but keeping the pattern as-is for now.

tracing/tracing.go
```diff
@@ -96,19 +94,19 @@ func (c Config) TracerProvider() (func(context.Context) error, error) {
 		semconv.SchemaURL,
 		semconv.ServiceNameKey.String(c.ServiceName),
 		semconv.ServiceNamespaceKey.String(c.ServiceNamespace),
-		semconv.ServiceVersionKey.String(os.Getenv("VERSION")),
+		semconv.ServiceVersionKey.String(c.ServiceVersion),
```
twessels-sh, line 97: So the `service.version` attribute was being read from the `VERSION` environment variable. But at least for the availability service, I'm not sure if this environment variable is ever set to anything other than the [chart default](https://github.com/spothero/availability/blob/main/basic-chart-config/availability/values.yaml#L74-L75), as I couldn't find any references to the variable in code, plus Jaeger traces have the chart default value of "local". I therefore changed the attribute value to use the `ServiceVersion` config value, which looks like it should be set when our Go services are built (at least it looks that way [in the availability service](https://github.com/spothero/availability/blob/main/cmd/availability.go#L19)).

grpc/client.go
theisenmark: This looks like a safe change to me 👍

theisenmark (approved):

tracing/tracing.go
```diff
@@ -46,22 +46,38 @@ const CorrelationIDCtxKey CorrelationIDCtxKeyType = iota
...
 // TracerProvider returns an OpenTelemetry TracerProvider configured to use
-// the Jaeger exporter that will send spans to the provided url. The returned
+// the OTLP exporter that will send spans to the provided url. The returned
 // TracerProvider will also use a Resource configured with all the information
 // about the application.
 func (c Config) TracerProvider() (func(context.Context) error, error) {
+	return CreateTracerProvider(c, getOtlpExporter)
+}
+
+func CreateTracerProvider(c Config, otlpExporterGetter OtlpExporterGetter) (func(context.Context) error, error) {
```
twessels-sh, line 80: The changes to `TracerProvider()` and the introduction of this method and `getOtlpExporter()` are necessitated by the 0.13% drop in test coverage from my [initial commit](https://github.com/spothero/tools/pull/239/commits/f0866e52682d9509a01152359cbffcd10031ddc7#diff-116b48d5813ab8b17a439ae27a88997ebf5f2c54c66a9172307b626582b94c10). I needed to introduce a way to mock a failure when creating the exporter. If there is a better way of doing this, let me know. 

Another option is to remove the "could not initialize OTEL exporter" log statement.

